PR #28, opened 2016-10-17
Fix stream_params() to correctly detect location

Suppaman:
stream_tweets() does not seem to correctly detect locations in the q argument.

It seems the first else statement in stream_params() is not capable of reaching {params <- list(locations = stream)} because the corresponding if statement checks is.integer() on the result of as.numeric() and therefore is (almost) always true.

If not mistaken, changing as.integer() to as.numeric() should fix it.


mkearney:
Thanks! If the tests fail on Travis for whatever reason, I will double check things later today!

PR #80, opened 2017-05-16
Support for trends/closest endpoint

bdilday:
Gets trends closest to a given latitude, longitude location as mentioned in https://github.com/mkearney/rtweet/issues/79


mkearney:
Beautiful! Thank you!

PR #89, opened 2017-06-06
Use svg version of travis badge

hadley:


mkearney:
Thank you @hadley for this and all your feedback! And this package wouldn't exist, for the record, without Adv-R. So thank you for that, too!

PR #90, opened 2017-06-23
Explain in documentation when this parameter is silently ignored

giocomai:
One can waste a lot of time troubleshooting what is going wrong. Giving this information in the documentation prevents that.


mkearney:
I'm all too familiar with the troubleshooting time sinks. Thanks for the documentation fix, @giocomai!

PR #94, opened 2017-07-02
typo

cderv:
This change is [Reviewable](https://reviewable.io/reviews/mkearney/rtweet/94)


R/tokens.R
```diff
@@ -5,7 +5,7 @@
 #'   to create and save an R user profile declaring the path to your
 #'   Twitter tokens. This allows Tokens to be instantly  [re]loaded
 #'   for future sessions. It also makes it easier to write teh card -
```
cderv: it says `write teh card`, seems not the right word. I understand `tech card` but not sure it is another typo

PR #116, opened 2017-10-04
In intro vignette: post_follow_user -> post_follow

mine-cetinkaya-rundel:
The vignette mentions the functions `post_follow_user` for following users, but the correct function for doing this seems to be `post_follow`. This PR simple updates these references.


mkearney:
good catch. Thanks @mine-cetinkaya-rundel !

PR #131, opened 2017-10-28
Update statuses.R

luke-a:
Doc correction

### DIFF
--- a/R/statuses.R
+++ b/R/statuses.R
@@ -1,7 +1,7 @@
 #' lookup_tweets
 #'
-#' @description Returns Twitter user data_frame object for
-#'   specified user_ids or screen_names.
+#' @description Returns Twitter status data_frame object for
+#'   specified statuses.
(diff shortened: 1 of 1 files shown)


luke-a:
Oops, I just saw the @description also needs polishing: "Returns Twitter user data_frame object for specified user_ids or screen_names." --> "Returns Twitter status data_frame object for specified status ids."

mkearney:
Thanks for this! I actually spent much of the past week or two revamping all of the documentation. I had already changed these files before seeing this request, but you did catch an incorrect reference to "user" (instead of "status") I missed. These kinds of pull requests are useful to me, though, so don't hesitate in the future (I'll hopefully be better about checking the PR before I start doing my own editing next time :)).

luke-a:
I think this was my 1st try of a PR. Glad it somehow worked out and you find it helpful for maintaining the package. :-)

PR #152, opened 2017-12-11
Fix for `get_timeline` ignoring `home` parameter

davidbody:
This fixes a bug where `get_timeline` was ignoring the `home` parameter.

I also added a regression test for this and fixed a typo in the `get_timeline` documentation.

### DIFF
--- a/R/timeline.R
+++ b/R/timeline.R
@@ -67,6 +67,7 @@ get_timeline <- function(user,
     user = user,
     n = n,
     max_id = max_id,
+    home = home,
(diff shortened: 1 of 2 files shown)


davidbody:
Perhaps it would make sense to have a `get_home_timeline` function and remove the `home` parameter from `get_timeline`. The Twitter API has separate endpoints for these.

This would be a breaking change in `rtweet`'s API, but that fact that `get_timeline` currently ignores `home` means such a change is unlikely to break any existing code.

PR #162, opened 2018-01-01
verify no spaces in usernames for get_favorites

tombuckley:
I was (admittedly, stupidly) calling get_favorites with a username that included a space in it. The get_favorites function returned rows that were favorited by my authenticating username, but then related them to the usernames with spaces in them. 

This was confusing to me. I added a simple check on the usernames to see if they have spaces, with a test. I also checked all tests and they passed. 

I totally understand this may be too basic/simple. Anyway, thanks for this package @mkearney!

### DIFF
--- a/R/favorites.R
+++ b/R/favorites.R
@@ -71,7 +71,7 @@ get_favorites_call <- function(user,
     token <- check_token(token)
   }
   ## check inputs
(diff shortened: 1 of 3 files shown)


mkearney:
This is great! Thank you!

PR #215, opened 2018-04-05
stream data now has a field for retweet counts, retweet_retweeted_count

thomas-keller:
Hey, this should in theory be pretty easy to merge. I just followed your code style. The field was already parsed, it just wasn't being used.

### DIFF
--- a/R/tweets_and_users.R
+++ b/R/tweets_and_users.R
@@ -421,6 +421,7 @@ statuscols_ <- function() {
     retweet_text = "retweet_text",
     retweet_created_at = "retweet_created_at",
     retweet_source = "retweet_source",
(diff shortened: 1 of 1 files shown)


mkearney:
I'll run some additional tests to be sure, but it appears you came up with the solution really fast! Thank you!

thomas-keller:
It's a bit silly, but I just noticed this so I thought I'd pass it along. The field I added is in reverse order to all the other retweet/favorite fields throughout the dataframe (the top-level retweet/favorite counts, quote_counts). Could you reverse the order so that retweet_favorite_count comes before retweet_retweeted_count in the dataframe for consistency?

PR #221, opened 2018-04-09
Fix source

thomas-keller:
This pull fixes the retweet_source and quoted_source columns so they use the parsed data, and uses your internal clean_source_ function to strip out the ungodly href garbage from the beginning of the source tags


mkearney:
Sorry this took so long. I got confused by the two pull requests and didn't want to override `full_text` stuff that integrated (however poorly) earlier :)

thomas-keller:
In my head I thought I was making things easier by submitting two patches, but I guess I just made it worse, huh?

Thanks for unraveling it.

PR #244, opened 2018-06-02
Update OAuth instructions for callback locking default change

mikechapple:
Twitter changed the default behavior of API apps to enable callback locking.  This breaks rtweet and causes a 403 error.  This update modifies the authentication vignette accordingly.


mkearney:
Beautiful! Thank you!

PR #259, opened 2018-06-20
Add open_in_browser() to view objects in browser

Kudusch:
Use open_in_browser() to generate and open links in the default browser
to view tweets, users or lists on twitter.com.

Sometimes it's useful to look a tweet as it appears on twitter.com and it can be cumbersome to generate the correct link manually. With open_in_browser() you can just pass a tweet, user or list and view them in the default browser. 

Warns if more than 10 links will be opened and uses utils::browseURL().

Contains function, documentation and addition to NAMESPACE.


mkearney:
I've been contemplating adding a `status_url` column in the parsed return object. This is the internal function I have lying around: https://github.com/mkearney/rtweet/blob/master/R/tweets_and_users.R#L670-L689. I like your idea of generating URLs for the user, the tweet, and/or the list.

+ `user_url` https://twitter.com/{screen_name}/
+ `status_url` https://twitter.com/{screen_name}/status/{status_id}
+ `list_url` https://twitter.com/{screen_name}/lists/{slug}

Before I proceed, though, I wonder if automating the opening of a browser (or 10 or 50 browser windows) to a Twitter URL violates Twitter's TOS?

PR #275, opened 2018-07-15
Fix typo in code example

xvrdm:
Fixing "acess" to "access" in "Authenticate via access token"


xvrdm:
BTW the "Authenticate via access token" is missing "consumer_secret" but it is present in the Rmd: probably pkgdown that needs to be rebuilt.

Thanks for the awesome package!

jonocarroll:
FYI, this persists in `README.md` -- needs to be re-knit.

Issue #164, opened 2018-01-07
`get_thread()` useful?


hrbrmstr:
I had a TODO to read a friend's recent tweet thread and did so easily in twitter app-proper.

Later on I saw some thread with a stable genius popup book and at the end the author @'d a bot that did a thread capture (which did not work, but meh).

We don't need apps or bots. We have `rtweet`!

Quick and dirty `get_thread()` with a `print()` method:

```r
#' Retrieve an authors tweet thread/rant from a starting status id
#'
#' @param first_status the first status id in the tweet thread
#' @param .timeline_history how much timeline history of the tweet author
#'        to pull to use when searching for the thread?
#' @return `list` with a data frame of the tweets and a text representation of the
#'         thread.
get_thread <- function(first_status, .timeline_history=3000) {

  require(rtweet, quietly=TRUE, warn.conflicts=FALSE)
  require(igraph, quietly=TRUE, warn.conflicts=FALSE)
  require(tidyverse, quietly=TRUE, warn.conflicts=FALSE)

  # get first status
  orig <- lookup_tweets(first_status)

  # grab the author's timeline to search
  author_timeline <- get_timeline(orig$screen_name, n=.timeline_history)

  # build a data frame containing from/to pairs (anything the author
  # replied to) that also includes the `first_status` id.
  filter(author_timeline,
         (status_id == first_status) |
           (reply_to_screen_name == orig$screen_name)) %>%
    select(status_id, reply_to_status_id) %>%
    graph_from_data_frame() -> g # build a graph from this

  # decompose the graph into unique subgraphs and return them to data frames
  decompose(g) %>% 
    map(as_data_frame) -> threads_dfs

  # find the thread with our `first_status` ids
  thread_df <- keep(threads_dfs, ~any(which(unique(unlist(.x, use.names=FALSE)) == first_status)))

  # BONUS: we get them in the order we need!
  thread_order <- discard(rev(unique(unlist(thread_df))), str_detect, "NA")

  # filter out the thread from the timeline corpus & sort it
  filter(author_timeline, status_id %in% pull(thread_df[[1]], from)) %>%
    mutate(status_id = factor(status_id, levels=thread_order)) %>%
    arrange(status_id) -> tweet_thread

  # extract the text and make it something readable
  pull(tweet_thread, text) %>%
    paste0(collapse="\n\n") -> thread_text

  # TODO: HTML version of ^^
  
  # make a small data structure to hold both of ^^ since I dislike attributes
  list(
[...]

PR #305, opened 2019-01-21
add functions for tweet embedding and threading

yonicd:


yonicd:
the threading function needs a leaf of a thread to run, which is a problem for ongoing threads. I also have a version the can start from the root or mid thread and continue to the last published leaf. Is it worth making the threading to be able to traverse in either direction?

llrs:
This PR might close #164. I think it makes sense to look up and down.

llrs:
Thanks @yonicd finally it can be merged

yonicd:
i totally forgot about this PR (2 yrs is a long time :) ). 

thanks for merging it!

PR #310, opened 2019-02-02
Adds auto_populate_reply_metadata

schnee:
https://github.com/mkearney/rtweet/issues/309

Adds `auto_populate_reply_metadata` which seems to increase the probability of Twitter threading a reply to a tweet generated by an account not associated with the token used by the rtweet instance. Without this, Twitter may or may not appropriately thread.


mkearney:
This is cool! Thank you for this!

PR #337, opened 2019-06-21
Bug in tweets_and_users.R when lookup_users returns multiple results with embedded data frams

jimeharrisjr:
Previous version failed when pulling large numbers of IDs with an error such as 

> Error in `[[<-.data.frame`(`*tmp*`, i, value = c(1434514262L, 1434582534L,  : 
  replacement has 10000 rows, data has 100


mkearney:
I'm closing this b/c I created a mess. Please submit pull request with your name added to updated DESCRIPTION as a contributor. The suggest code change was implemented. Thanks!

PR #349, opened 2019-08-19
fix magrittr version

Enchufa2:
Simple fix: it's 1.5, not 1.5.0. For R, they are equal. But this causes issues in other systems (e.g., RPM packaging).


mkearney:
Good to know! Thanks!

PR #372, opened 2019-11-12
swap unroll_connections inputs/outputs to not mangle UTF-8 on windows

knapply:
I've been working on building twitter networks elsewhere and had already laid a foundation before realizing that rtweet now has some built-in functionality, so I've got some lessons-learned that may be of interest.

I changed `unroll_connections()` so that it doesn't mangle UTF-8 data on Windows (`std::string` pains); building networks such as user-to-hashtag will now be safe for Windows folks if that's something desired in the future.

The algorithm is a bit simpler than the previous one and skips the `std::vectors`. My benchmarks are showing that it's >= 3x faster.

### DIFF
--- a/man/bearer_token.Rd
+++ b/man/bearer_token.Rd
@@ -39,21 +39,21 @@ authentication method
 verified_user_tweets <- search_tweets("filter:verified", n = 30000, token = bearer_token())
 
 ## get followers (user == app)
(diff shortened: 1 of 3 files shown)


knapply:
closing after Rcpp was dropped in fdf8a175beb8380c25a235575df55d0aa4e380fe

PR #375, opened 2019-12-12
Fix Twitter Premium API Max Requests in search_tweets.R

kevintaylor:
Update search_tweets such that the search_fullarchive function supports max 500 tweets per request, which is the limit for the Twitter premium API. Fixing this bug is important because it can double or more the cost of using the Twitter premium API.


kevintaylor:
@mkearney is there something else I can do to get this fix ready for you?

tguerra:
It's not yet fixed for the `search_30day()` function with premium API which only returns max 6k tweets (12 requests).
I tested it with d110184.

`packageVersion("rtweet")
‘0.7.0.9000’`

PR #417, opened 2020-05-05
fix parsing of path to token during auto-auth

### LINKED ISSUE #413: cannot open compressed file '/Users/xx.rds', probable reason 'Permission denied
```
> library(rtweet)
> rt <- search_tweets(
+   "#rstats", n = 18000, include_rts = FALSE
+ )
Requesting token on behalf of user...
Waiting for authentication in browser...
Press Esc/Ctrl + C to abort
Authentication complete.
Error in gzfile(file, mode) : cannot open the connection
In addition: Warning message:
In gzfile(file, mode) :
  cannot open compressed file '/Users/clayton1.rds', probable reason 'Permission denied'
```

Hello I am unable to authenticate my token either by the browser as shown above, or with an app and credentials. How do I point the save to somewhere else than "/Users/clayton1.rds' where I clearly don't have permissions

MichaelChirico:
Closes #413 

The regex to do `no_ext` is wrong. See here:

https://regex101.com/r/12j8Xl/1

We could fix the regex, but I think the approach here is simpler & relies on existing `tools::file_ext` to find the file extension.

Alternatively, I think `no_ext` would work on `basename(x)`:

```
paste_before_ext = function(x) {
  f = basename(x)
  file.path(dirname(x), no_ext(f), p, only_ext(f))
}
```

### DIFF
--- a/R/tokens.R
+++ b/R/tokens.R
@@ -225,31 +225,10 @@ create_token_ <- function(app = "mytwitterapp",
   token
 }
 
-
-has_ext <- function(x) {
-  stopifnot(length(x) == 1L)
-  x <- basename(x)
-  grepl("[[:alnum:]]{1,}\\.[[:alpha:]]{1,}$", x)
-}
-
-only_ext <- function(x) {
-  if (has_ext(x)) {
-    gsub(".*(?=\\.)", "", x, perl = TRUE)
-  } else {
-    ""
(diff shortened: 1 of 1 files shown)


MichaelChirico:
it looks like CI failed because I'm writing to a fork & don't get the right permissions there

PR #472, opened 2021-02-15
No deletion of stream folder

### LINKED ISSUE #294: stream_tweets2() does not save the output
### Problem

`stream_tweets2()` deletes the .json output . I can see it creates a temporary folder while it's running, but the folder is deleted when the functions stops.

`stream_tweets2(q = c(-86.6874, 29.6644,-85.1626, 30.5846), dir = 'tweet2', timeout = 5, include_rts = F, parse = F)`

### Session info

> sessionInfo()
R version 3.5.1 (2018-07-02)
Platform: x86_64-apple-darwin15.6.0 (64-bit)
Running under: macOS Sierra 10.12.4

Other attached packages:
 [1] rtweet_0.6.8    rvest_0.3.2     xml2_1.2.0      forcats_0.3.0   stringr_1.3.1   dplyr_0.7.6     purrr_0.2.5    
 [8] readr_1.1.1     tidyr_0.8.1     tibble_1.4.2    ggplot2_3.0.0   tidyverse_1.2.1

llrs:
Should close #294. Thanks @Kwazao!

Although a better path it might be just deprecate the function stream_tweets2. As noted on the comments it is not the best place to test functions. But now that people use it at least make it work as well as possible.

### DIFF
--- a/R/stream.R
+++ b/R/stream.R
@@ -526,7 +526,7 @@ stream_tweets2 <- function(..., dir = NULL, append = FALSE) {
   pat <- paste0(file_name, "\\-[[:digit:]]{1,}\\.json$")
   jsons <- list.files(dir, pattern = pat, full.names = TRUE)
   file_name <- paste0(dir, ".json")
-  unlink(dir, recursive = TRUE)
+  #unlink(dir, recursive = TRUE)
   if (!parse) {
     return(invisible())
   }


llrs:
Forgot to tag you: @ropensci/rtweet This PR is a simple change made some time ago

Issue #446, opened 2020-09-20
get_timelines broken


climbthemt:
### Problem

Error on pulling tweets using 'get_timelines'
First noticed problem 4 months ago, hoped it would get a bug fix.  

### Expected behavior

My existing code normally pulls n=xxx tweets. Now at the end of the pull there is an error and no tweets are pulled.
Normally I'd get 200 tweets from user "AOC" stored in the tmls object.
My existing code has been debugged for months and was working fine.

### Reproduce the problem

tmls <- get_timelines(c("AOC"), n = 200)

>Error in isFALSE(dots[["parse"]]) : could not find function "isFALSE"

### rtweet version

‘0.7.0’

### Session info

R version 3.4.4 (2018-03-15)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Ubuntu 18.04.5 LTS

Matrix products: default
BLAS: /usr/lib/x86_64-linux-gnu/atlas/libblas.so.3.10.3
LAPACK: /usr/lib/x86_64-linux-gnu/atlas/liblapack.so.3.10.3

locale:
 [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C               LC_TIME=en_US.UTF-8        LC_COLLATE=en_US.UTF-8    
 [5] LC_MONETARY=en_US.UTF-8    LC_MESSAGES=en_US.UTF-8    LC_PAPER=en_US.UTF-8       LC_NAME=C                 
 [9] LC_ADDRESS=C               LC_TELEPHONE=C             LC_MEASUREMENT=en_US.UTF-8 LC_IDENTIFICATION=C       

attached base packages:
[1] parallel  stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
 [1] doMC_1.3.6       iterators_1.0.12 foreach_1.5.0    forcats_0.5.0    stringr_1.4.0    purrr_0.3.4      readr_1.3.1     
 [8] tibble_3.0.3     tidyverse_1.3.0  tidytext_0.2.5   tidyr_1.1.2      lubridate_1.7.9  dplyr_1.0.2      ggplot2_3.3.2   
[15] rtweet_0.7.0    

loaded via a namespace (and not attached):
 [1] tidyselect_1.1.0  haven_2.3.1       lattice_0.20-41   colorspace_1.4-1  vctrs_0.3.4       generics_0.0.2   
[...]

PR #474, opened 2021-02-18
Update R version

llrs:
Updates the minimum version required because the package uses isFalse identified in #385. I think the suggestion of #385 although good will make it harder if I had to support versions of 7 years ago.

Closes #385 and #446.

@sckott @mkearney


sckott:
I agree to bump the minimum R version

PR #475, opened 2021-02-18
Remove Rcpp and document again

llrs:
Removes Rcpp and documents should fix the note from CRAN: "Namespace in Imports field not imported from: ‘Rcpp’".
If other things get merged before I will need to re-document again once this gets merged.

@sckott @mkearney


sckott:
Looks good

PR #482, opened 2021-02-21
Pkgdown improvements

### LINKED ISSUE #435: rtweet.info is down
If there are no plans to get it backup, perhaps update the links to the official ropensci docs: http://docs.ropensci.org/rtweet

hadley:
* removes outdated copy of website in `docs/` (not needed since now built by ROpenSci)
* updates website urls (Fixes #435)
* start dividing reference page into useful categories


llrs:
Thanks, had some changes on the pkgdown configuration coming, for instance using the ropensci template, but not sure how the current pkgdown website is build (@sckott ?) Thanks for dividing the references into topics!

sckott:
@llrs you can check on docs builds here https://dev.ropensci.org/ 

> not sure how the current pkgdown website is build 

what do you mean?

llrs:
The current _pkgdown.yml documentation doesn't use the ropensci template but the documentation does use it, to me it looks like the file is ignored. Should I keep the Google tracking number?

sckott:
@llrs the docs are docs.ropensci.org/pkg are only built from the default branch (main/master) - so I don't think it's ignored, you just won't see it until this is merged. I think we have tracking? @jeroen ?

llrs:
This is out of scope of this PR currently we don't see the default pkgdown template on https://docs.ropensci.org/rtweet/ as the configuration file uses. Is the configuration file modified by the build system? I will need to reread the documentation

I suppose you have tracking enabled but this comes from the previous site/url, so I'm not sure if it is okay to provide another cookie/tracking or should I remove it. Let me know and I'll adjust accordingly.

sckott:
I don't really know the answers to these questions. Let's wait for jeroen

jeroen:
Our build system calls `pkgdown::build_site()` except that it overrides the template to be [rotemplate](https://github.com/ropensci-org/rotemplate). So indeed if you want, you could remove the `template` section from the _pkgdown.yml (including the GA key), because it is redundant (but it doesn't hurt).

PR #487, opened 2021-02-24
Fix two small buglets in new token code

hadley:


llrs:
Wasn't sure if the global assignment was or wasn't needed. I suppose that relying on the standard environment solving algorithm from R works.

PR #488, opened 2021-02-24
Use inheritParams to reduce duplication in docs

hadley:
Starting with token and parse arguments

@llrs please take a look at the new docs found in `lookup_users()` and let me know if there's anything I could to make the documentation more clear. 

Once this is merged, I can convert all the docs to use markdown.

### DIFF
--- a/R/bearer_token.R
+++ b/R/bearer_token.R
@@ -32,8 +32,7 @@ create_bearer_token <- function(token = NULL) {
 #' Convert default token into bearer token for application only (user-free)
 #' authentication method
 #'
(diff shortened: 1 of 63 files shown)


llrs:
You left in several functions the `#' @param user` as it was and in some they match with the lookup description. Is this intentional? Otherwise it looks fine. I don't merge yet in case it wasn't.

hadley:
Sorry, I don't follow. I'm probably missing something obvious, but I don't think I touched any `@param user` descriptions?

llrs:
Oh, sorry got confused between user and users.

PR #492, opened 2021-02-24
Don't attach package in try_require()

hadley:
Because there's no need, and generally packages shouldn't change the user's search path.


jeroen:
Thanks for your contributions @hadley! I'll give you write access to the repo, so you can help maintain the package.

PR #494, opened 2021-02-24
Improve media handling

llrs:
Improve upon the media and alt text new features. 
Basically testing before making any request to twitter API and provide a reasonable error message. 
@sckott I also exported a new function `post_destroy` to reduce the work of `post_tweet` and provided a minimal test for them.


R/post.R
```diff
@@ -110,7 +94,7 @@ post_tweet <- function(status = "my first rtweet #rstats",
 
     ## send request
     r <- TWIT(get = FALSE, url, token)
-
+    httr::stop_for_status(r)
```
hadley, line 97: I'd suggest leaving these off, so we can tackle in one swoop in `TWIT_get()`.

sckott:
@llrs no need to ping me anymore if hadley's involved 😉

R/post.R
```diff
@@ -80,23 +80,7 @@ post_tweet <- function(status = "my first rtweet #rstats",
...
-    r <- TWIT(get = FALSE, url, token)
-
-    ## if it didn't work return message
-    if (r$status_code != 200) {
-      return(httr::content(r))
-    }
-    ## if it did, print message and silently return response object
-    message("your tweet has been deleted!")
-    return(invisible(r))
+    return(post_destroy(destroy_id))
```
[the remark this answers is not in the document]
llrs: Added the .Deprecated message. However, the documentation of `post_destroy` is  in `post_tweet`. So I haven't deleted the argument documentation. Do you think it is better in its own page?

R/post.R
```diff
@@ -148,13 +135,28 @@ post_tweet <- function(status = "my first rtweet #rstats",
...
   }
-  if (length(status) > 1) {
-    stop("can only post one status at a time",
-         call. = FALSE)
-  }
+
 
   ## media if provided
   if (!is.null(media)) {
+    if (length(media) > 4) {
```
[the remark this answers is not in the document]
llrs: Created, also moved all the media related functions to its own file

R/post.R
```diff
@@ -203,6 +204,38 @@ is_tweet_length <- function(.x, n = 280) {
   nchar(.x) <= n
 }
 
+# equivalent to tools::file_ext
```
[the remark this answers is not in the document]
llrs: I was just moving around the function, but now it is replaced by tools::file_ext

R/post.R
llrs: Removed

R/post.R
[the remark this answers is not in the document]
hadley: I'd suggest keeping them as is to avoid conflicts with #498.

R/post.R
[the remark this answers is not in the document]
hadley: I think you're better off adding a lifecycle dependency since the user experience is much better (i.e. only one warning every 8 hours) unless you have a strong reason not to add a lightweight dependency.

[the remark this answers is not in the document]
llrs: I don't see the annoyance of getting a message each time a users use post_tweet to eliminate a tweet too much. They can move to the new function if this bothers them too much. But I'll replace it by the lifecycle version. Regarding user experience, we can also use cli for the message and notifications to users to make the package more user friendly.

llrs:
Some PR changed the DESCRIPTION so now there's a conflict. But @hadley take a look if you find the changes reasonable and think this can be merged.

R/post.R
```diff
@@ -65,6 +65,7 @@
 #' }
 #' @family post
 #' @aliases post_status
+#' @importFrom lifecycle deprecate_warn
```
hadley, line 68: You don't need to import this since you're using it with `::`

R/post.R
```diff
@@ -307,6 +286,58 @@ check_chunked_media_status = function(finalize_data, token, rurl) {
   }
 }
 
+
+#' @importFrom tools file_ext
```
hadley, line 290: Similarly, no need to import this.

R/post.R
```diff
@@ -307,6 +286,58 @@ check_chunked_media_status = function(finalize_data, token, rurl) {
...
+  if (any(nchar(alt_text) > 1000)) {
+    stop("Alt text cannot be bigger than 1000 characters.", call. = TRUE)
+  }
+}
+
+
+
+#' @export
+#' @rdname post_tweet
+post_destroy <- function(destroy_id, token = NULL) {
```
hadley, line 317: Let's wait on merging this until you can use `TWIT_post()`.

R/post.R
llrs: Does it have any downsides?

hadley: No downsides — you should only either use `::` or `@importFrom`, not both.

R/post.R
hadley: I've changed my mind — I now think it makes sense to merge this before #506.

tests/testthat/test-post_tweet.R
```diff
@@ -0,0 +1,8 @@
+test_that("post_tweet works", {
+  skip_if_offline()
+  skip_on_cran()
+  
+  rt <- expect_message(post_tweet("test"), "your tweet has been posted!")
```
hadley, line 5: I just realised this is probably a bad idea — I don't really want to be sending tweets from my twitter account just by running the tests. We should skip this until we have a twitter account just for testing.

llrs: You got funny replies :D But that means that post_destroy didn't work well, so it was a good test. Anyway this also highlights the need of using/managing different tokens

PR #496, opened 2021-02-25
Eliminate home_user references

### LINKED ISSUE #489: Eliminate home_user()
* In `get_my_timeline()`, instead use [`home_timeline()`](https://developer.twitter.com/en/docs/twitter-api/v1/tweets/timelines/api-reference/get-statuses-home_timeline) which doesn't require a user name
* In `lists_users()`, just use `NULL` (since if no user supplied, [endpoint](https://developer.twitter.com/en/docs/twitter-api/v1/accounts-and-users/create-manage-lists/api-reference/get-lists-list) returns authenticated user)
* Similarly, in `post_list_add_one()`, `post_list_destroy()`, `post_list_create_all()`, `post_list_destroy_all()`, use `NULL` for user id

llrs:
Simplified the timeline process and should close #489 but:
- I removed the vectorization of users to add_list
- I need to check if this works when the token is provided with acces_token & access_secret and not validated with just browser because some functions need the user screen name and I get it via token$credential$screen_name


llrs:
I don't know how to continue on this. To add to the list one needs the owner_id or screen_name. If the user is authenticated just with the browser the screen name and the user_id is added to the token but when the authentication is with the access_secret and access_token then we don't know who the user is.
@hadley Do you think a modification version of `authenticating_user_name` could be called in `create_token` ? 

Also the post_list needs to  be splited in several functions too and simplified.

hadley:
I think it's ok to let this sit for a little while — it'll be easier to tackle once we've done more refactoring elsewhere.

Issue #468, opened 2021-01-30
Support for Academic Product Track?


nsingh23:
Hi,

Im just wondering if there are any plans to support the "Academic Product Track" version that Twitter has recently released. Ive been trying to use the current rtweet version, but it throws an error. Note that the Academic Track does not have a devenv and I've set mine to NULL.

https://dev.to/twitterdev/getting-historical-tweets-using-the-full-archive-search-endpoint-1agp

"v2 full-archive search endpoint" - 

![image](https://user-images.githubusercontent.com/75192524/106346479-7cdd9180-6285-11eb-852d-24f4c7b504e9.png)

llrs:
There have been a long hiatus on the maintenance of the package. But hope it can support APIv2 in the near term, a PR would speed the process.

schochastics:
I have written a function that supports the "Academic Product Track" ([gist](https://gist.github.com/schochastics/1ff42c0211916d73fc98ba8ad0dcb261)). It is not very opinionated but does the job for now.

nsingh23:
Thank you so much David!. The pagination is a lifesaver! 

Much appreciated!

JBGruber:
Thanks @schochastics, this is a fantastic start and I can confirm it works for me!

``` r
source("https://gist.githubusercontent.com/schochastics/1ff42c0211916d73fc98ba8ad0dcb261/raw/040422b5e1378ef4c30150d4927a3991f53bc922/get_tweets.R")

df <- get_tweets(
  q = "#rstats", 
  n = 10, 
  start_time = "2010-01-01T00:00:00Z", 
  end_time = "2010-02-01T00:00:00Z",
  token = "redacted"
)

tibble::as_tibble(df$data)
#> # A tibble: 10 x 12
#>    entities$mentio… $annotations $hashtags conversation_id possibly_sensit…
#>    <list>           <list>       <list>    <chr>           <lgl>           
#>  1 <df[,3] [1 × 3]> <df[,5] [2 … <df[,3] … 8467623082      FALSE           
#>  2 <NULL>           <df[,5] [2 … <df[,3] … 8467441206      FALSE           
#>  3 <df[,3] [1 × 3]> <NULL>       <df[,3] … 8462375244      FALSE           
#>  4 <df[,3] [1 × 3]> <NULL>       <df[,3] … 8453279722      FALSE           
#>  5 <df[,3] [1 × 3]> <NULL>       <df[,3] … 8431308826      FALSE           
[...]

PR #497, opened 2021-02-25
Start to pull out TWIT_get() function

hadley:
@llrs this work isn't completed, but I think it's in a good place for you to take a look to make sure you understand how `TWIT_get()` works. I'd recommend that you start by reading the source code for `TWIT_get()`, asking me about any techniques that I've used that you don't yet understand. Then take a look at the functions that I've rewritten.


llrs:
I think I understood all the code but not why you write it that way. I understand the `withr::local_options` but don't understand why you use it instead of the `options` and `on.exit`. 
Another thing I don't understand well is about `stop_for_status`. You don't use `httr::stop_for_status` but deferred to `check_status`, but there is also the `warn_for_twitter_status` (used many times with httr::GET) and `check_status_code` (seems to be used with TWIT_POST). Why this choice? Should some of these be eliminated ? (Perhaps not as part of this PR)

Also the scipen option used on the package are a bit inconsistent but the max found is 15 (not sure if it will matter or is just an error). But I got some post with an id of 19 numbers, perhaps a bigger number is reasonable? 

It might be too soon but the API is moving to v2. Perhaps a new argument to `TWIT_GET` with default version = "1.1", could be set. So that when new api v2 endpoints start to be implemented we don't need to go back and modify this function.

hadley:
I use `with::local_options()` because it's simpler, and makes it harder to make mistakes (like forgetting `add = TRUE`).

Totally agree that we'll need to rationalise all the ways of checking status — my plan is to get rid of `warn_for_twitter_status()` and `check_status_code()` as I rewrite those functions (and as we eliminate use of bare httr calls)

In general, I don't think it's a good idea to speculatively add arguments to functions. It's too hard to predict what the right API will be until you have some example code. But I do think `TWIT_get()` and

llrs:
I was thinking mainly on the academic research product track, highly requested lately (#468), is on API v2. But certainly we can wait till we get the package on shape.

hadley:
@llrs I think it's the right approach, I just don't think it's the right time — we can add an easy way to use v2 once we start implementing those functions. Anyway, since you're on board, I'm going to merge these commits. If they cause merge conflicts in any of your other PRs please let me know and I can take a look.

PR #500, opened 2021-02-26
More translations to TWIT_get()

hadley:
Focussing on existing uses of `httr::GET()`

There are about five left, which I'll take a look at tomorrow. Probably won't be able to get them all in this pass since some appear to be tangled up in pagination code.


hadley:
All done — note that I deleted the `suggested_*()` functions since they haven't worked since June 2019.

hadley:
@llrs I don't think this needs a close read, mostly sharing so that you can see what I'm working on.

llrs (approved):

PR #503, opened 2021-02-26
Start refactoring pagination

hadley:
Built on top of #500


hadley:
I think as an interim step I'll not worry about the code that waits for rate limits, and then add that all back in as a separate step.

Next, need to implement `TWIT_paginate_by_cursor()`.

hadley:
@llrs I'm not sure it's possible to review such a massive PR, but you can look at the `TWIT_paginate_` functions to see how things work. After this is merged, I'll still need to restore the rate limit waits, and continue using `TWIT_paginate_cursor()` in more places.

R/lists_statuses.R
```diff
@@ -82,5 +50,15 @@ lists_statuses_ <- function(list_id = NULL,
     tweet_mode = "extended"
   )
 
-  TWIT_get(token, "lists/statuses", params)
+  results <- TWIT_paginate_max_id(token, "lists/statuses", params,
+    get_max_id = function(x) x$id_str,
+    page_size = 200,
```
llrs, line 55: On the documentation we said something else: 

> n optional Specifies the number of results to retrieve per "page.".

Might be better to change the documentation to the max, number of tweets to get. 
Also about documentation, this function does not have any examples.

R/next_cursor.R
```diff
@@ -92,182 +92,17 @@ next_cursor.response <- function(x) {
...
-  )
-  max_id(.x)
-}
-
-#' @export
-max_id.response <- function(.x) {
-  .x <- from_js(.x)
-  max_id(.x)
+max_id <- function(.x) {
+  stop("max_id() is deprecated; rtweet now handles pagination for you")
```
llrs, line 99: This doesn't make use of `lifecycle::deprecate_warn`, not sure how it was used previously so not sure if this error each time is more user friendly than a warning.

R/favorites.R
```diff
@@ -42,133 +37,78 @@ get_favorites <- function(user,
...
-    fav <- tweets_with_users(fav)
+    results <- tweets_with_users(results)
+    results$favorited_by <- user
+  }
+  
+  results
+}
+
+TWIT_paginate_max_id <- function(token, query, params, 
+                                 get_max_id,
```
llrs, line 88: A comment about why get_max_id parameter is needed and a explaining why should a function be passed will make maintenance easier, imho.

R/premium.R
```diff
@@ -111,147 +76,103 @@ format_from_to_date <- function(x = NULL) {
...
+    fromDate = format_from_to_date(fromDate),
+    toDate = format_from_to_date(toDate),
+    max_id = max_id,
+  )
+  
+  api <- paste0("search/tweets/", product, "/", env_name)
+  result <- TWIT_paginate_max_id(token, api, params,
+    get_max_id = function(x) x$statuses$id_str,
+    n = n,
+    page_size = 100,
```
llrs, line 148: In permium search isn't it the maxResults parameter used instead of page_size? Also it is 500 the maximum: https://developer.twitter.com/en/docs/twitter-api/premium/search-api/api-reference/premium-search#Pagination
Several users complained that their queries using rtweet became more expensive due to not  fully use the API. This also applies to fullarchive.

R/search_tweets.R
```diff
@@ -187,247 +187,68 @@ search_tweets <- function(q, n = 100,
...
-    count <- n
-  } else {
-    count <- 100
-  }
+  type <- match.arg(type)
+  
   ## validate query length–char count might not always be same here as with 
   ## Twitter, so set this to 600 and let Twitter reject others
   if (nchar(q) > 600) {
     stop("q cannot exceed 500 characters.", call. = FALSE)
```
llrs, line 228: There's a mismatch between the error message and the nchar used to trigger it.

llrs (approved):

R/friends.R
```diff
@@ -76,123 +76,38 @@ get_friends <- function(users,
...
+  
+  results <- TWIT_paginate_cursor(token, "friends/ids", params, 
+    page_size = 5000,
+    n = n
+  )
+  
+  if (parse) {
+    results <- tibble::tibble(
+      user = user,
+      ids = unlist(lapply(results, function(x) x$ids))
```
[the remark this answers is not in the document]
hadley: I've never used `getElement()` before, but looking at the source code it seems like it's a wrapper around `slot()` and `[[`, and since we're not using S4 here, I don't think it's needed.

R/get-my-timeline.R
```diff
@@ -35,16 +35,14 @@ get_my_timeline <- function(n = 100,
                             token = NULL,
                             ...) {
 
-  user <- home_user()
-
-  args <- list(
-    user = user,
+  get_timeline(
+    user = home_user(),
```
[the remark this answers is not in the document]
hadley: Yeah, we'll work on that in that PR.

R/search_tweets.R
hadley: This is unrelated to the thrust of this PR, so it's better to resolve elsewhere.

R/followers.R
```diff
@@ -85,156 +85,52 @@ get_followers <- function(user, n = 5000,
...
-  if (has_name_(f, "next_cursor_str")) {
-    ## next cursor
-    nc <- f[["next_cursor_str"]]
-    if (is.null(nc)) return("0")
-    return(nc)
+TWIT_paginate_cursor <- function(token, query, params, n = 5000, page_size = 5000) {
+  params$count <- page_size
+  
+  # TODO: consider if its worth using fastmap::faststack() here
+  results <- list()
```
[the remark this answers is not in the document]
hadley: That won't make any appreciable difference to performance.

R/next_cursor.R
hadley: I think this was mostly for internal use; we can bring it back later if we discover folks need it.

R/premium.R
hadley: Good catch, I'll fix.

llrs (approved):

PR #505, opened 2021-02-26
Improve rate_limit() interface

hadley:
I've implemented the infrastructure needed to automatically wait until the rate limit is reset, but I'm not sure what the interface should be. I think it should require explicit opt-in because 15 minutes is a long time to wait, but it seems like it would be nice if there was a global option?


hadley:
And paginate methods need to recognise (via signal?) so they can terminate early

PR #506, opened 2021-02-26
Split post.R up into multiple files

hadley:
I've made 2 failed attempts to switch to `TWIT_post()` — I think the problem is that the `post.R` file is just too large and it's hard to see how all the functions related. I've made a first attempt at splitting the functions into separate files.


hadley:
Closing this since of all the merge commits — I still think it's a reasonable next step, but I'm not just quite ready for it yet.

llrs:
Sorry I missed to review this PR.

PR #507, opened 2021-02-27
Improve the maximum n warnings

resulumit:
This pull request improves the maximum n warnings, making them consistent with each other. One of these warnings had a typo. All four warnings had been written in different ways.


hadley:
Can you please file this as an issue instead? We're currently in the middle of a big refactoring of these functions, so this would be better done after we're finished. I'd also like to think more about whether these warnings are necessary at all (should they just be left to the twitter API?) or if they remain, whether they'd be better as errors.

PR #511, opened 2021-02-27
Chunk mp4 + tests

dieghernan:
`mp4` is always chunked now. Also tests for media (and extra files for testing) added


tests/testthat/test-post.R
```diff
@@ -1,12 +1,42 @@
...
+
+  ## test-chunked
+  id <- upload_media_to_twitter(path, chunk_size = fs - 3)
+  expect_type(id, "character")
+
+  # mp4
+  path <- test_path("tweet.mp4")
+  fs <- file.size(path)
+
+  ## test-unchunked
```
hadley, line 34: Isn't this always chunked?

tests/testthat/test-post.R
```diff
@@ -1,12 +1,42 @@
 test_that("upload_media_to_twitter() can handle small file", {
   skip_on_cran()
-  
-  id <- upload_media_to_twitter(test_path("tweet.gif"))
+
+  # gif
+  path <- test_path("tweet.gif")
```
hadley, line 5: I pulling `path` out into a separate variable makes the test harder to understand because it introduces indirection.

tests/testthat/test-post.R
```diff
@@ -1,12 +1,42 @@
...
+  fs <- file.size(path)
```
hadley, line 6: Similarly, I think using a small constant makes it easier to reason about.

tests/testthat/test-post.R
```diff
@@ -1,12 +1,42 @@
...
   # after it's been uploaded
-  
-  id <- upload_media_to_twitter(test_path("tweet.gif"), chunk_size = 1024)
+
+  ## test-chunked
+  id <- upload_media_to_twitter(path, chunk_size = fs - 3)
+  expect_type(id, "character")
+
+  # png
+  path <- test_path("tweet.png")
```
hadley, line 19: Unless you think there's something materially different in the treatment of pngs and gifs, I think you should omit this test as it doesn't actually touch any more lines of code in this package.

hadley (commented):
Thanks! I've made a few comments on your test to improve readability.

tests/testthat/test-post.R
dieghernan: Agree, I was just double-checking

tests/testthat/test-post.R
dieghernan: Ok, I am leaving just gif and removing also `tweet.png` file

dieghernan:
I think I addressed all your comments, now there are just 3 tests (`gif` chunked and unchunked and `mp4` test for forced chunking). I also removed the `png` as it is not used any more.

Thanks

hadley:
Looks great, thanks!

Issue #485, opened 2021-02-22
rstats2twitter access_token & access_secret 


llrs:
Basically we might not be able to recover/have access to the access_token & access_secret of rstats2rtweet. For regular users it doesn't mater as they authenticate with their own username or use their account. 

To test the application they are needed, ideally a premium account to test premium API's limits and endpoints and academic product and others. I could use my personal token but that would lock the package with my account. The package is under the guiding/protecting hands of [rOpenSci](https://ropensci.org) and specially of @sckott. It would be better if they were able to renew or address anything related to the Twitter app if the maintainer (me or the next one) isn't available.

@andypiper What is the best route to have a new one for testing? Create a new user and request for a new app? I created a new user app_rtweet in case we need to create a new token just for the package. 

Thanks

llrs:
I requested a developer account for a user just for testing & developing the package [app_rtweet](https://twitter.com/app_rtweet). I requested a team permission mentioning rOpensci organization. Hope it gets approved soon

llrs:
Closing this as we went to create a new account/tokens. Just for completeness [rtweet_test](https://twitter.com/rtweet_test) is the current default used for tests  while [app_rtweet](https://twitter.com/app_rtweet) was approved as a set up a team developer account.

Issue #469, opened 2021-01-31
Support multiple tokens


llrs:
### Problem

If multiple tokens are present only the one selected by the enviroment variable TWITTER_PAT is used.
So users need to remember where are they stored and which one is which. 

I think a better system would be to list all the tokens available. (This way it could also warn if a new token for an existing app is being created)

To list all the tokens 

```
tk <- list.files(path = "~", pattern = ".rtweet_token*", full.names = TRUE, all.files = TRUE)
```

Also the token app name can be recovered by: `token$app$appname` so that the user can choose from which account it consumes the API calls.

hadley:
This might also be a good place to consider the problem revealed by #528 — how can we allow contributors to run "safe" calls (i.e. GETs) with their personal twitter account, without accidentally run "dangerous" calls (i.e. POSTs) that might change their account, send tweets, DMs, etc.

hadley:
Hmmmmm, maybe the tests should default to using a bearer token?

hadley:
Having spent a little time with `bearer_token()` I think it might be worth a holistic review of authentication. I think there are three main forms of authentication:

* app: collecting data
* user: casual user, take actions on behalf of user
* bot: take actions on behalf of app

And each uses a slightly different set of data from the "keys and tokens" page:

* app: bearer token
* user: api key & secret (+ interactive user auth)
* bot: api key & secret, access token & secret

<img width="451" alt="Screenshot 2021-03-05 at 1 00 46 PM" src="https://user-images.githubusercontent.com/4196/110163592-da4d8c80-7db5-11eb-8dc3-2aaff9a02335.png">

The only scenario that requires caching is "user", because the interactive auth generates a couple of pieces of data that we want to preserve over time. But all scenarios want to be able to change the default token for the session. I wonder if it might be better to set up the auth around these these scenarios:

* `auth_app(bearer)`
* `auth_user(api_key, api_secret)`
* `auth_bot(api_key, api_secret, access_key, access_secret)`

And then provide some way to set the default token. We'd still use caching for the default user token, but for anything else you'd call the appropriate auth function and then set the default:

```R
auth_set_default(auth_app("AXACSFASDFA--332342"))
auth_set_default(auth_bot(...))
auth_set_default(loadRDS("my-cache.Rds"))
```

@llrs do you think that would cover for your use case?

llrs:
[...]

PR #512, opened 2021-02-27
Use twitter account specifically for tweeting

hadley:
To avoid a repeat of https://twitter.com/hadleywickham/status/1365661250269683713 😆


hadley:
Fixes #485

hadley:
@llrs I've merged this just so I can keep on working. I think we should plan for a comprehensive review of this code as part of #469.

PR #521, opened 2021-02-28
Eliminate home_user() in favour of api_screen_name()

### LINKED ISSUE #489: Eliminate home_user()

hadley:
@llrs I made this PR since I've made #496 impossible to merge with merge conflicts

Fixes #489


llrs (approved):
I think it would be nice if it also returned the user_id not only the screen_name. I think that for list I needed to set both to make work that endpoints.

hadley:
Why do you need both? Either is unambiguous.

llrs:
I might be wrong and only one was needed. Anyway I'll find out

Issue #510, opened 2021-02-27
Consistent pagination interface


hadley:
## Cursors

Add `next_cursor` attribute to outptut (of single inputs). Should pass `cursor, `retryratelimit` and `verbose` arguments on to `TWIT_paginate_cursor()`. 

* [x] `direct_messages()`
* [x] `get_followers()`
* [x] `get_friends()`
* [x] `lists_members()`
* [x] `lists_memberships()`
* [x] `lists_subscribers()`
* [x] `lists_subscriptions()`

## `max_id`

Add `max_id` attribute to outptut (of single inputs). Should pass `since_id`, `max_id`, `retryratelimit` and `verbose` on `TWIT_paginate_max_id()`.

* [x] `get_favorites()`
* [x] `get_my_timeline()`
* [x] `get_timeline()`
* [x] `get_mentions()`
* [x] `lists_statuses()`
* [x] `search_tweets()`

## Other

* [ ] `search_users()` — uses `page` paramter
* [x] `lookup_users()` — `TWIT_paginate_chunked()`
* [x] `lookup_tweets()`  — `TWIT_paginate_chunked()`

hadley:
Consider `sleep_on_rate_limit` instead of `retryratelimit`. Would need to deprecate functions already using `retryratelimit`.

hadley:
@llrs do you have any thoughts on argument name? I think `sleep_on_rate_limit` is slightly easier to read and a more accurate description of what it does, but I don't feel that strongly about it. (But I do only want to make the change once, since it's a bunch of functions that need this as an argument.)

hadley:
Common documentation in progress in #554

llrs:
I think it might be slightly easier to understand wait_on_rate_limit, but it is still long. (wait_on_limit?)

hadley:
Cursor plan:

* Add `rtweet_cursor`  attribute to result of `TWIT_paginate_cursor`
* Ensure parsing preserves/restores the attribute
* Re-work `next_cursor()` following `max_id()`/`since_id()`
* Ensure all functions have necessary arguments

hadley:
I'm going to leave `my_friendships()` and `lookup_friendships()` as is as there doesn't appear to be any indication that they're actually used anywhere (e.g. https://github.com/search?l=R&q=my_friendships&type=Code).

hadley:
I'm going to close now; `search_users()` uses its own unique pagination style so while we certainly could make consistent with everything else, I don't think it's particularly high priority.

PR #523, opened 2021-02-28
Drastically simplify rate_limit()

hadley:
* [x] Default rate limit error needs to give more details; should give limits and time needed to wait
* [x] In paginated functions, hitting a rate limit should terminate early with a warning, but still return all data retrieved so far
* [x] Optionally (and with global option) should be able to opt-in to instead waiting


hadley:
@llrs you don't need to read the code, but I'd appreciate you trying out the user experience with `search_tweets()` and `get_followers()`. The documentation and messaging should make it clear what's going on, so please let me know if you have any confusion.

I need to stop working on rtweet and get back to Mastering Shiny, so the pace of contributions should be much slower this week.

llrs:
Not sure it is working, I tried `rate_limit("/lists/list")` but it failed when trying to initialize an empty tibble: Error: Input must be a vector, not NULL.
From the user experience, previously one could use `rate_limit("lookup_users")` and it would use the right endpoint. But that can be added later.

Also I tried `st <- search_tweets("#rstats", n = 1000, retryonratelimit = TRUE)` to see what would happen and got a:
> Rate limit exceeded for Twitter endpoint 'search/tweets' . Waiting for refresh in  X mins /


So from other functions seems to work well, not sure the user needs to know endpoint perhaps it could be shortened to 

> Rate limit exceeded. Waiting for refresh in X mins.

llrs:
Many thanks for all the heavy lifting so far! 
I will look at the other issues left and slowly work on them.

hadley:
Good catch, I fixed that.  Note that #510 tracks the application of these principles to every function.

llrs (commented):
LGTM

PR #525, opened 2021-02-28
Use TWIT_post in post.R

hadley:
@llrs I think this should be my final contribution for at least a week. When combined with #523, this now means that all requests that can use `TWIT_get()` or `TWIT_post()`  do use them.


llrs (approved):
LGTM

llrs (approved):

Issue #480, opened 2021-02-21
Add function to list muted accounts


hadley:
eg. https://developer.twitter.com/en/docs/twitter-api/v1/accounts-and-users/mute-block-report-users/api-reference/get-mutes-users-ids

llrs:
Thanks! Yes, indeed is something I found missing. Fixing #467 would also help if people over-muted

hadley:
I think the problem is that `post_follow()` is trying to do too much — it calls one of four different API endpoints depending on the combination of arguments used. I'd be happy to do a PR to fix this if you wanted.

llrs:
Yes, PR are more than welcomed, if you have time send it please. Just not sure when I'll merge them. As  currently I can't use the github actions to test the package, and I would like to prioritize fixing the long-standing open issues following #471

hadley:
I think the most important thing is to get the tests working again, but I don't see an issue that tracks that?

llrs:
Indeed! Maybe I should create one. There are a couple of things with tests:
1) I don't have access to the keys and token of the rstats2twitter app provided by rtweet (which is used for the test). They are not in Travis or GHA. I just sent today an email to @mkearney, asking him and explaining the situation, if I don't get any reply in a couple of weeks I will go ahead and provide a new token (Probably would get in touch with Twitter developers to make sure this is ok).

2) I'm kind of new contributing to this repository and my PR are first reviewed by @sckott. I provided a minor improvement in #478. As I didn't get any response on that front I halted a bit. But I created another branch (improve-tests) for a more through review of tests.

hadley:
I'll wait until you're done with #527, before working on this issue, but I've been starting to think about the interface a little bit. We might as well tackle blocking at the same time in the same fashion, so I was thinking of the following six functions:

```R
user_mute()          # POST mutes/users/create
user_unmute()        # POST mutes/users/destroy
user_block()         # POST blocks/create
user_unblock()       # POST blocks/destroy

user_list_muted()    # GET blocks/ids
user_list_blocked()  # GET mutes/users/ids
```

I think it makes the most sense to start the function names with the "type" of thing (user, tweet, list, message) because I think   that's the most obvious part — you might not know exactly what type of operation you want to apply, but it seems likely that you'll know what to apply it to. Then we follow with a verb, matching how people normally talk about these operations. 

[...]

PR #527, opened 2021-03-03
Add tests from examples

llrs:
Re introduce some changes from PR #494 and adds tests for several functions from the examples. In some cases those examples were needed to update (users do not longer exists).


R/followers.R
```diff
@@ -55,7 +41,7 @@ get_followers <- function(user, n = 5000,
                           page = lifecycle::deprecated()) {
   
   if (lifecycle::is_present(page)) {
-    lifecycle::deprecate_warn("1.0.0", "get_followers(page)", "get_followers(cursor)")
+    lifecycle::deprecate_warn("1.0.0", "get_followers(page = )", "get_followers(cursor = )")
```
hadley, line 44: If the existing syntax doesn't work for you, it's a sign that we need to require lifecycle 1.0.0 in the description, and you need to update.

R/post_destroy.R
```diff
@@ -0,0 +1,19 @@
...
+#' Deletes a status of user's profile.
+#' @inheritParams post_tweet
+#' @export
+#' @examples
+#' \dontrun{
+#' pt <- post_tweet()
+#' crt <- httr::content(pt)
+#' post_destroy(crt$id_str)
+#' }
+post_destroy <- function(destroy_id, token = NULL) {
```
hadley, line 12: We should probably briefly contemplate whether this would be better called `tweet-destroy()` or `status_destroy()` — I think the existing `post_` prefix confusingly refers to the HTTP method, not a twitter "post" (i.e. a tweet).

tests/testthat/test-lookup_users.R
```diff
@@ -0,0 +1,10 @@
+test_that("lookup_users works", {
+  users <- c(
+    "potus", "hillaryclinton", "realdonaldtrump",
+    "fivethirtyeight", "cnn", "espn", "twitter"
+  )
+  usr_df <- lookup_users(users)
+  expect_s3_class(usr_df, "tbl")
+  expect_equal(nrow(usr_df), 6)
+  expect_equal(ncol(usr_df), 91)
```
hadley, line 9: I think it's risky to hard code these numbers because they might fail for reasons outside of our control (i.e. a user gets banned, or the Twitter api changes what fields it returns).

tests/testthat/test-post.R
```diff
@@ -24,3 +24,14 @@ test_that("can set alt text", {
...
+
+
+test_that("post_tweet works", {
+  skip_if_offline()
+  skip_on_cran()
+  msg <- paste("test", Sys.time()) # To avoid having duplicated status
+  rt <- expect_message(post_tweet(msg), "your tweet has been posted!")
+  crt <- httr::content(rt)
+  rt <- post_destroy(as.character(crt$id))
+  expect_error(post_destroy(as.character(crt$id_str)))
```
hadley, line 36: I'd suggest putting this in a separate test since it's no longer testing `post_tweet()`

tests/testthat/test-get_friends.R
```diff
@@ -0,0 +1,4 @@
+test_that("get_friends works", {
+  djt <- get_friends("ropensci")
+  expect_s3_class(djt, "tbl")
```
hadley, line 3: It's mildly more correct to use `tbl_df` here.

tests/testthat/test-post.R
llrs: I don't understand, do you suggest to have it in a different file (test-post_destroy) or to split these 4 lines in two tests, one for post_tweet another for post_destroy? If I split the test I wouldn't be able to know which status/tweet I want to delete and then I would end up posting a new one and deleting it again which is what currently it is doing.

tests/testthat/test-lookup_users.R
llrs: Precisely, I would like to know if the API changes or the user we are using for testing and the examples is banned (we could always use the user specific for rtweet to avoid last problem).

R/post_destroy.R
llrs: We are discussing this on #480

R/followers.R
llrs: Ok, I removed this change, but I'll remove the dependency to lifecycle later on in my opinion it is not worth it. I prefer to notify the user each time, specially considering th next release is a major release.

hadley: I think that is something that’s worse discussing more - lifecycle provides a consistent way to deprecate functions and it’s designed to be a lightweight dependency.

hadley: Maybe my earlier explanation wasn’t very good. I can fix the problem on master on Monday so you don’t need to worry about it in this PR.

tests/testthat/test-post.R
hadley: You’d still deletion test the tweet in the creation test, but you’d only test the double deletion in the deletion. It is hard to keep creation and deletion tests disentangled in this sort of code but I think it’s worth striving for as much as possible. (I was just commenting on that one specific line, not the test in general)

tests/testthat/test-lookup_users.R
hadley: Ok, makes sense. The main thing is to acknowledge and accept the risk.

llrs:
I haven't included the test for post_list as there is a bug on post_list that needs to be fixed.
Also not sure how post_follow(retweets = FALSE) should work, tried to make a test for it but didn't manage to pass the checks consistently. Leaving here the code:

```r
test_that("post_follow works without retweets", {
  # First we need to follow then we can disable retweets
  # pf <- post_follow("BarackObama")
  pf <- post_follow("BarackObama", retweets = FALSE)
  expect_equal(httr::status_code(pf), 200L)
  pf <- post_follow("BarackObama", destroy = TRUE)
  expect_equal(httr::status_code(pf), 200L)
})
```

Several code lines/functions can be deleted: `go_get_var` and `any_recursive` are not used.
Also there isn't currently a function to delete direct message so I get some messages each time the pacakge is tested. 

@hadley I think that this PR might be ready to get merged. Let me know if there is anything else to improve on the tests.

.github/workflows/pr-commands.yaml
```diff
@@ -34,6 +34,8 @@ jobs:
     runs-on: macOS-latest
     env:
       GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
+      RTWEET_ACCESS_TOKEN: ${{ secrets.RTWEET_ACCESS_TOKEN }}
```
hadley, line 37: I don't _think_ these are needed for the PR commands workflow?

R/post-tweet.R
```diff
@@ -227,3 +224,25 @@ wait_for_chunked_media <- function(resp, media_id, token = NULL) {
 
   invisible()
 }
+
+check_media <- function(media, alt_text) {
+  media_type <- tools::file_ext(media)
+  if (length(media) > 4) {
+    stop("At most 4 images per plot can be uploaded.", call. = FALSE)
```
hadley, line 231: Now that we import rlang, you could simplify these to 

```suggestion
    abort("At most 4 images per plot can be uploaded.")
```

tests/testthat/test-direct_messages.R
```diff
@@ -0,0 +1,13 @@
...
+  expect_length(dm, 2L)
+})
+
+test_that("direct_messages_received works", {
+  expect_error(direct_messages_received(), "no longer exists")
+})
+
+test_that("direct_messages_sent works", {
+  expect_error(direct_messages_sent(), "no longer exists")
+})
```
hadley, line 13:
```suggestion
test_that("old functions give informative errors", {
  expect_error(direct_messages_received(), "no longer exists")
  expect_error(direct_messages_sent(), "no longer exists")
})
```

tests/testthat/test-get_followers.R
```diff
@@ -0,0 +1,5 @@
+test_that("get_followers works", {
+  kfc <- get_followers("KFC")
+  expect_s3_class(kfc, "tbl")
+  expect_equal(nrow(kfc), 5000)
```
hadley, line 4: I'd suggest setting `n = 100` or similar, just to keep the tests as fast as possible

tests/testthat/test-network_data.R
```diff
@@ -0,0 +1,17 @@
+test_that("network_data works", {
+    rstats <- search_tweets("#rstats", n = 20)
```
hadley, line 2: Looks like the indentation is wider than usual here

tests/testthat/test-post.R
```diff
@@ -24,3 +24,14 @@ test_that("can set alt text", {
   id <- upload_media_to_twitter(test_path("tweet.gif"), alt_text = "A bird tweeting")
   expect_type(id, "character")
 })
+
+
+test_that("post_tweet works", {
+  skip_if_offline()
+  skip_on_cran()
```
hadley, line 31:
```suggestion
```

tests/testthat/test-post_follow.R
```diff
@@ -0,0 +1,16 @@
...
+  expect_equal(httr::status_code(pf), 200L)
+})
+
+test_that("post_destroy works", {
+  pf <- post_follow("BarackObama", destroy = TRUE)
+  expect_equal(httr::status_code(pf), 200L)
+})
+
+test_that("post_follow works without notifications", {
+  pf <- post_follow("BarackObama", notify = FALSE)
```
hadley, line 12: `notify = FALSE` is the default?

tests/testthat/test-post_message.R
```diff
@@ -0,0 +1,6 @@
+test_that("post_message works", {
+  rt <- expect_message(post_message(paste("Testing", Sys.time()), 
+                                    user = "Lluis_Revilla"),
+                       "your DM has been posted!")
+  expect_equal(httr::status_code(rt), 200L)
```
hadley, line 5:
```suggestion
  rt <- expect_message(
    post_message(paste("Testing", Sys.time()), user = "Lluis_Revilla"),
    "your DM has been posted!"
  )
  expect_equal(httr::status_code(rt), 200L)
```

tests/testthat/test-search_tweets.R
```diff
@@ -15,3 +15,16 @@ test_that("gives useful errors", {
   expect_snapshot(search_tweets(c(1:10), verbose = FALSE), error = TRUE)
   expect_snapshot(search_tweets("stats", type = "all"), error = TRUE)
 })
+test_that("search_tweets2 returns tweets data and latlng", {
+  skip_on_cran()
+  skip_if_offline()
+
```
hadley, line 21:
```suggestion
```

tests/testthat/test-search_tweets.R
```diff
@@ -15,3 +15,16 @@ test_that("gives useful errors", {
...
+  df <- search_tweets2(c("#rstats", "open science"), n = 50)
+  expect_s3_class(df, "tbl_df")
+  expect_true(nrow(df) > 25) # should almost always be true
+  
+  # can extract lat_lng
+  ll <- lat_lng(df)
+  expect_equal(c("lat", "lng") %in% names(ll), c(TRUE, TRUE))
```
hadley, line 28: I think you can drop this

tests/testthat/test-search_tweets.R
```diff
@@ -15,3 +15,16 @@ test_that("gives useful errors", {
   expect_snapshot(search_tweets(c(1:10), verbose = FALSE), error = TRUE)
   expect_snapshot(search_tweets("stats", type = "all"), error = TRUE)
 })
+test_that("search_tweets2 returns tweets data and latlng", {
```
hadley, line 18:
```suggestion
test_that("search_tweets2 can search for multiple queries", {
```

hadley (commented):
Overall, looks great. I just gave you a bunch of very nitpicky feedback.

R/post-tweet.R
llrs: We don't use classes of errors so I'll stick with base stop.

tests/testthat/test-get_followers.R
llrs: Yeah, did so with others but missed this one.

R/post-tweet.R
hadley: `abort()` also gives you a better default for `call.` (so you don't forget) and generates more informative tracebacks.

PR #528, opened 2021-03-04
Factor out hardcoded versioning in API calls

alexpghayes:
@hadley @llrs I've started work on this but am having trouble testing code locally. 35 tests are being skipped with `Reason: Auth not available`. I do however have my credentials saved locally and `get_token()` successfully finds those credentials and turns them into a token object.

### DIFF
--- a/R/access_level.R
+++ b/R/access_level.R
@@ -1,5 +1,5 @@
 api_access_level <- function(token = NULL) {
-  r <- TWIT_get(token, "account/settings", parse = FALSE)
+  r <- TWIT_get(token, "/1.1/account/settings", parse = FALSE)
 
   if ("headers" %in% names(r) && "x-access-level" %in% names(r$headers)) {
     r$headers$`x-access-level`
--- a/R/as_screenname.R
+++ b/R/as_screenname.R
@@ -97,6 +97,6 @@ api_screen_name <- function(token = NULL) {
     skip_status = TRUE,
     include_email = FALSE
   )
-  r <- TWIT_get(token, "account/verify_credentials", params)
+  r <- TWIT_get(token, "/1.1/account/verify_credentials", params)
   r$screen_name
 }
--- a/R/collections.R
+++ b/R/collections.R
@@ -36,7 +36,7 @@ lookup_collections <- function(id, n = 200,
     count = n,
     ...
   )
-  TWIT_get(token, "collections/entries", params, parse = parse)
+  TWIT_get(token, "/1.1/collections/entries", params, parse = parse)
 }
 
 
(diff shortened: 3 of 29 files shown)


llrs:
The token now is saved in a new location under a new name: `~/.cache/rtweet/auth.rds`. You probably need to move/copy your token there. Or create a new token with the dev version that will be saved there. However, as I wanted to provide a smooth transition, could you provide where is your current token, and which appname is token you obtain with get_token ?

hadley:
@llrs I think the problem is more likely to be the logic here — https://github.com/ropensci/rtweet/blob/master/R/tokens.R#L31-L37. @alexpghayes do the tests work when run interactively in the console? If so, the problem is the deliberate decision not to use your person token for tests, since they can potentially send tweets. I'm not sure how we can resolve this — maybe we should only require a special token for the tests of `POST` end points?

hadley:
Also note that the tests are automatically skipped in this PR; that's because @alexpghayes isn't a contributor and hence the repo secrets aren't available to him. I don't currently see any way around this; I think this is just the type of project where all contributions have to be verified locally by a maintainer.

hadley:
I converted a couple more that @alexpghayes missed; the test coverage for the POST functions is pretty low right now, but I think it's ok to accept that this PR might break them (or they might already be broken) and we'll fix them as we increase test coverage.

llrs:
So, as I said, the current logic does not check that there is a valid token on the previous default location. As far as I know the existing test if they post something they should also undo the action (unless the later fails). I'm working on another PR to increase the test coverage #527, so this PR can be merged once more tests are added there and we know how much code does this break (if any).

alexpghayes:
@llrs Can you cc me when you merge #527 so that I can start working on next steps?

llrs:
Of course @alexpghayes, I'll let you know. Hope to finish it this weekend, or on Monday the latest. Thanks for the PR!

alexpghayes:
Okay, just tried rebasing to resolve the conflict. I don't rebase often so confirmation that I haven't done a git nasty would be appreciated. My process was 

1. pull `ropensci/master` into `my_local/master` to get your updates, 
2. rebase `my_local/versioned_TWIT_method` onto `my_local/master`, and 
3. force push `my_local/versioned_TWIT_method` to `alexpghayes/versioned_TWIT_method`.
 
 The force push is what I'm most concerned about.

hadley:
Looks fine to me 😄

llrs:
Besides a weird initial no file changes on github everything look good! 
And all the tests passed :fireworks. Merged. Thanks!

PR #530, opened 2021-03-05
Start simplifying bearer token interface

### LINKED ISSUE #387: Confusion about bearer tokens
I'm trying to understand bearer tokens better by using two tokens that I have for different applications.

I have two tokens, `token1` and `token2`, that are clearly different:

``` r
> identical(token1, token2)
[1] FALSE
```

However, the corresponding bearer tokens are identical. When and why is this the case?

``` r
> bearer1 <- bearer_token(token1)
> bearer2 <- bearer_token(token2)
> 
> 
> identical(bearer1, bearer2)
[1] TRUE
```

hadley:
The bearer token is no longer cached; it's not an expensive operation to retrieve it, it's app specific, and the user should be storing in a variable anyway. A bearer token is (nearly) equivalent to having the client app and secret, so I've made it harder to expose by accident with a new print method, and made it more difficult to use rtweets default app.

The tests currently fail because invalidate_bearer() doesn't work, probably because I don't understand the asymmetry between getting an access token and invalidating an access token.

Fixes #387


hadley:
Still no luck getting bearer token invalidation to work 😞. But the existing `invalidate_bearer()` function doesn't work either, so it's not like this PR has made things worse.

Issue #514, opened 2021-02-27
Eliminate skip_on_cran() and skip_if_offline() from tests


hadley:
* `skip_on_cran()` will happen automatically because all tests are skipped when auth isn't available
* I think `skip_if_offline()` is only needed for CRAN. It can't really happen for GitHub actions, and I think it's fine for the tests to fail when testing locally with no internet.

PR #533, opened 2021-03-09
Test gardening

hadley:
I've removed the now unneeded skips (#514) and renamed/reorganised the test files so that (e.g.) `tests/testthat/test-foo.R` tests the code in `R/foo.R`. This convention is easy to understand (previously we didn't have any good advice on how to organise your test files) and makes `usethis::use_test()`, `devtools::test_file()`, `devtools::test_coverage_file()` work.

@llrs it would be good to review/merge this quickly since it touches many test files and merge conflicts will be hard to resolve. Note that the tests are failing because I think we've activated some secondary rate limit on following and unfollowing users. We'll have to think more about how to test this.


llrs (commented):
I don't understand the logic behind naming the files. If I remember correctly testthat recommendations is to use test- and the name of the file with underscore (test-get_my_timeline) but some test files are all with - like test-get-my-timeline and some mix them like test-bearer_token.

hadley:
The logic is to match the files in `R/` — so the problem is that the files in `R/` are currently not consistent. Do you want me to rename them to consistently use `-`?

The problem is that I ran a bunch of tests in rapid succession and I think I tripped some bot threshold that doesn't go through the usual rate limit system. We might want to explore setting up the tests that modify data to only run once every 24 hours.

PR #535, opened 2021-03-14
Add geolocation feature to post_tweet

### LINKED ISSUE #501: [New feature] Add geolocation to tweets
Hi 

I have seen a modification or **rtweet** that allows to geolocate a tweet and I would like it to be implemented on the package:

https://github.com/mattkerlogue/narrowbotR/blob/34d491c4a626f9651965d64b65c7399cfc67e511/R/post_geo_tweet.R#L33-L36

See https://twitter.com/narrowbotR for real output

The implementation would be really easy IMO. I have prepared and succesfully tested a small code, (see Diff):

*Branch edited and rebased to latest changes*
https://github.com/ropensci/rtweet/compare/master...dieghernan:geotag-tweet

I could send a PR if you think it is something you may want to include.  I have seen that @llrs and @hadley both have been working hard last week on this, so maybe it could be a good moment to include a small new feature.

Twitter API Reference: https://developer.twitter.com/en/docs/twitter-api/v1/tweets/post-and-engage/api-reference/post-statuses-update

dieghernan:
closes #501 

I think this is ready for review, full API specification on:

https://developer.twitter.com/en/docs/twitter-api/v1/tweets/post-and-engage/api-reference/post-statuses-update

### DIFF
--- a/tests/testthat/test-post-tweet.R
+++ b/tests/testthat/test-post-tweet.R
@@ -31,6 +31,14 @@ test_that("post_tweet works", {
 })
 
 test_that("post_tweet geolocated works", {
+  # Test errors
+  msg <- paste("test geolocated error", Sys.time()) # To avoid having duplicated status
+  expect_error(post_tweet(msg, lat = "x", long = 0))
+  expect_error(post_tweet(msg, lat = 0, long = "x"))
+  expect_error(post_tweet(msg, lat = 91, long = 0))
+  expect_error(post_tweet(msg, lat = 0, long = 181))
+  expect_error(post_tweet(msg, lat = 0, long = 0, display_coordinates = "error"))
+  
   # Test geolocated tweet
   msg <- paste("test geolocated", Sys.time()) # To avoid having duplicated status
   rt <- expect_message(post_tweet(msg, lat = -36.811784, long = 174.792657), "your tweet has been posted!")
--- a/tests/testthat/test-post-tweet.R
+++ b/tests/testthat/test-post-tweet.R
@@ -29,3 +29,19 @@ test_that("post_tweet works", {
   rt <- expect_message(post_destroy(crt$id_str), "your tweet has been deleted!")
   expect_equal(httr::status_code(rt), 200L)
 })
+
+test_that("post_tweet geolocated works", {
+  # Test geolocated tweet
+  msg <- paste("test geolocated", Sys.time()) # To avoid having duplicated status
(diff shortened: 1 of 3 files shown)


llrs:
Sorry, forgot to check this PR, When I set the display_coordinates = TRUE I don't see anything on my mobile app different from display_coordinates = FALSE. Probably because on the testing user the geolocalization has not been enable. The documentation should mention that users need to set as TRUE the geo_enabled parameter (not sure where this is, user>configuration probably). Could you look this up and update the doc?

I added some checks on the function, so you'll need to pull this.

dieghernan:
Hi @llrs, you are right about `display_coordinates`, I tried to find the difference but I couldn't, however I implemented it as it is referred on the original documentation of the API (do you prefer to remove it?).

I would add some commits to the PR on the next days with the changes you proposed,

Best

dieghernan:
@llrs I have improved the docs in line with your suggestion. I also included some validity controls on the new parameters (should be numeric or logical , also explained on the docs, an on a valid range).

I updated also the related tests, I think I am hitting all the lines now. 

Best

hadley:
This needs a bullet in NEWS.md too

tests/testthat/test-post-tweet.R
```diff
@@ -29,3 +29,27 @@ test_that("post_tweet works", {
   rt <- expect_message(post_destroy(crt$id_str), "your tweet has been deleted!")
   expect_equal(httr::status_code(rt), 200L)
 })
+
+test_that("post_tweet geolocated works", {
+  # Test errors
+  msg <- paste("test geolocated error", Sys.time()) # To avoid having duplicated status
+  expect_error(post_tweet(msg, lat = "x", long = 0))
```
hadley, line 36: It's better to use `expect_snapshot()` for this sort of test because (a) that ensures you're generating the correct error and (b) it's allow the reviewer to see the exact text of the generated error.

dieghernan: Hi @hadley , thanks for your feedback. Should I send a new PR, as this has been already merged?

 I would need some time as I am not familiar with `expect_snapshot` and friends.

Regards

hadley: Yes, that's a good idea.

dieghernan: See #541 , I would appreciate further guidance if needed. Regards

Issue #524, opened 2021-02-28
Link to twitter api docs from every function


hadley:

hadley:
Fixed by #538.

PR #538, opened 2021-03-21
Add API references

llrs:
Added links on the documentation to the Twitter pages.

(Also added myself and Hadley as authors of the package).


hadley:
Fixes #524

R/post-favorite.R
```diff
@@ -14,6 +14,9 @@
 #' }
 #' @family post
 #' @export
+#' @references 
+#' Create: <https://developer.twitter.com/en/docs/twitter-api/v1/tweets/post-and-engage/api-reference/post-favorites-create>
```
hadley, line 18: You might want to make this a bulleted list.

llrs: Well, I hope that once we agree on the naming convention we will split some of these functions where multiple endpoints are used from the same function, so I'll wait till then to add these cosmetic touches.

PR #540, opened 2021-03-21
More consistent interface for user id vs screen name

### LINKED ISSUE #120: Mentions of deleted accounts cause confusion in tweets data
This is relatively minor. I found that mentions of accounts that have been deleted cause some discrepancies in the tweets dataframes. The "text" variable will still display the deleted account, but it's not displayed in the remaining variables where it normally would be (e.g., "reply_to_user_id", "mentions_user_id", etc.). 

I assume this discrepancy comes directly from the API, but if there were a good place to just mention it in the documentation that could be helpful. Took me a little bit to figure out what was going on.

The discrepancy can be found easily in some of my recent tweets: get_timeline("llsigerson").

### LINKED ISSUE #502: Handle user consistently
* All functions should allow either username or id (as string or double)
* One function to check if input is ok
* Helper for adding appropriate parameter to API call
* Determine if `as_screenname` and `as_userid` can be eliminated

hadley:
Since the assumption is that a string containing only numbers is already an id, there's no need for `as_userid()` and I deprecated it. 

Fixes #502. Fixes #120.


R/user_id.R
```diff
@@ -0,0 +1,78 @@
+#' Mark a user id as a screen name
+#' 
+#' @description 
+#' There are two ways to identify a twitter user: a screen name (e.g.
+#' "justinbieber") or a user identifier (e.g. "27260086"). User identifiers
+#' are 64-bit integers which can't be represented natively by any base R data
+#' structures, so in most cases rtweet returns them as strings. This introduces
+#' an ambiguity, because user names can also consist solely of numbers 
+#' (e.g. "123456"). You can resolve this ambiguity by using `as_screenname()` 
+#' to a character vector a screen name.
```
llrs, line 10: I think this section can be made more easier to understand. I would delete this sentence: "User identifiers are 64-bit integers which can't be represented natively by any base R data structures, so in most cases rtweet returns them as strings."
First, I hope we can return them consistently as strings (I think this is the case already), second to the user it doesn't matter what/how R works all important here, imho, is how to get rtweet to work. If you want to mention it I would make a note or place it at the end. 

Also it doesn't have a `#' @return` documentation even if trivial it will be helpful.

R/user_id.R
```diff
@@ -0,0 +1,78 @@
...
+
+user_type <- function(x, arg_name = "user") {
+  if (is.numeric(x)) {
+    "user_id"
+  } else if (is.character(x)) {
+    if (inherits(x, "rtweet_screen_name")) {
+      # needed for purely numeric screen names
+      "screen_name"
+    } else if (all(grepl("^[0-9]+$", x))) {
+      "user_id"
```
llrs, line 61: Shouldn't we check the size too with "^[0-9]{64}$" ? This would eliminate some false results as `user_type("123")` being user_id or `user_type("1244564654456546465464646412445646544565464654646464124456465445654646546464641244564654456546465464646412445646544565464654646464")`. May be worth also to leave a comment with a link from where this magical 64 comes from.

R/user_id.R
```diff
@@ -0,0 +1,78 @@
...
+#' @export
+#' @usage NULL
+as_userid <- function(x) {
+  lifecycle::deprecate_warn("1.0.0", "as_userid()")
+  x
+}
+
+user_type <- function(x, arg_name = "user") {
+  if (is.numeric(x)) {
+    "user_id"
```
llrs, line 55: On the documentation of as_screen we say most user_ids are returned as strings. Are all the screen_names returned as strings too? Because this assumes so. Couldn't find an example or a case but might be worth adding a comment.

tests/testthat/test-user_id.R
```diff
@@ -0,0 +1,16 @@
...
+
+test_that("user_type handles simple cases", {
+  expect_equal(user_type("hadleywickham"), "screen_name")
+  
+  expect_equal(user_type(123), "user_id")
+  expect_equal(user_type(bit64::as.integer64(123)), "user_id")
+  expect_equal(user_type("123"), "user_id")
+  expect_equal(user_type(as_screenname("123")), "screen_name")
+  
+  expect_snapshot(user_type(TRUE), error = TRUE)
```
llrs, line 15: Not sure why the snapshot and not a more easy to read `expect_error` here. 

Also might be worth to add the two extreme cases I mentioned in a different comment.

llrs (approved):
Great improvement, now it is much easier to follow how it is decided if a character is a screen name or a user id. Left some comments about some minor details.

R/user_id.R
hadley: I think it's worth noting that user ids look like numbers, but you can't store them in R as numbers. But I agree that comes too soon and most users won't care, so I'll move it later and make it a parenthetical.

R/user_id.R
hadley: I don't think it's worth putting too much magic in here, although it might be better to construct this test using `bit64::as.integer64()` since that's what we really care about.

llrs: Adding length check will reduce the potential errors for many users (specially those dealing with bots or newly created users which didn't set a user name), if you prefer use bit64 to check this.

hadley: Maybe we should annotate rtweet results so that that screen names are always marked as screen names and this becomes less of a problem?

llrs: Not sure how easy this would be, and if the user will see any benefit but I think it might be harder to implement, while checking this here the benefits are clearer and the change minimal.

hadley: It's a bad idea to sometimes declare that a character vector is a vector of screen names and sometimes its ids based on some relatively subtle feature of the string — that makes the overall behaviour of the system harder to predict and to debug when things go wrong. 

If you think it's likely that users will have vector of screen names containing only numbers it is better to solve the problem upstream, where the vector is created.

hadley: (Using `bit64::as.integer64()` doesn't work because it seems to use a very permissive parser, so I think it's best to leave this code as is)

PR #542, opened 2021-03-24
Overhaul auth mechanism to support multiple tokens/auth mechanisms

### LINKED ISSUE #387: Confusion about bearer tokens

### LINKED ISSUE #469: Support multiple tokens

hadley:
Fixes #469. Fixes #387.

I think this code is mostly done (although I do need to test it thoroughly). Most of the remaining work is to carefully deprecate the existing auth systems so that old code continues to work, but you are steered towards the new system.


hadley:
To do:

* [ ] Test 
* [ ] Update `token` param documentation
* [ ] Deprecate `get_token()`/`get_tokens()` in favour of `auth_get()`
* [ ] Deprecate `create_token()` in favour of `auth_type_app()` and `auth_type_bot()`. Suggest `auth_save()` if `set_renv()` is `TRUE`.
* [ ] Update `auth.Rmd`
* [ ] Update section in README
* [ ] Check if `FAQ.Rmd` still needed

llrs:
I can help with the several tokens I use. (Sorry I haven't reviewed the other PR, but I haven't found time to look at it in detail yet).

R/auth.R
```diff
@@ -0,0 +1,237 @@
...
+#' @description 
+#' There are three ways that you can authenticate with the twitter API:
+#' 
+#' * `auth_user()` interactively authenticates an existing twitter user. 
+#'   This form is most appropriate if you want rtweet to control your
+#'   twitter account.
+#'   
+#' * `auth_app()` authenticates as a twitter application. An application can't 
+#'    perform actions (i.e. it can't tweet) but otherwise has generally higher 
+#'    rate limits (i.e. you can do more searches) generally higher. This form is
```
alexpghayes, line 12:
```suggestion
#'    rate limits (i.e. you can do more searches). This form is 
```

hadley: Thanks; added manually since I've made a bunch of other changes.

hadley:
@llrs I think this is in a good place to try out / review now. It's a big PR so I'd suggest starting in `auth.Rmd` which should hopefully explain how all the bits fit together

@alexpghayes would also be useful to get your eyes on `auth.Rmd` to see if this resolves your questions about using bearer tokens.

R/auth.R
```diff
@@ -0,0 +1,292 @@
...
+#' @param name Cache name to use.
+#' @export
+#' @examples 
+#' \dontrun{
+#' # save app auth for use in other sessions
+#' auth <- rtweet_app()
+#' auth_save(auth, "my-app")
+#' 
+#' # later, in a different session...
+#' auth_us("my-app")
```
alexpghayes, line 139:
```suggestion
#' auth_as("my-app")
```

vignettes/auth.Rmd
```diff
@@ -12,104 +12,76 @@ vignette: >
...
+Assuming that's good, you can now run `auth_save()` to save your authentication object so that you can use it again in the future:
+
+```{r, eval = FALSE}
+auth_save(auth, "your-app-name")
 ```
 
-That's it!
+## Use
+
+Now, in a future session, you can use your saved auth by calling `auth_as()`:
```
alexpghayes, line 83: The language here makes it sound a little bit like you can only save app auth, since you recover an auth via the app name. I'm also curious how this works when all the auths come from the same app. In theory I could have one Twitter project with one app and use that same application name for all three types of authentication, if my understanding is correct. How does selecting amongst those strategies work here since there is only a single app name for three auth strategies in that case (again, subject to me understanding the Twitter API correctly)?

hadley: You can save any auth object. The name is arbitrary, so presumably you'd use some suffix to distinguish them. Maybe I need to make that more clear?

alexpghayes: I think that would help. I think the instructions about how to use `rtweet` are fairly clear, but using the same app name for all three types of authentication might lead people in poor mental models of Twitter side authentication.

alexpghayes:
Is there a function that will display the available saved auth methods so that you can select between them / set a default one? In the auth vignette I currently see tools for creating and saving auth, but am less clear on how to manage multiple auth types at once.

hadley:
@alexpghayes what do you mean by manage multiple auth types? Do you mean multiple saved auth objects?

alexpghayes:
> @alexpghayes what do you mean by manage multiple auth types? Do you mean multiple saved auth objects?

I mean whether you're authorizing as bot, app, or user. For example, suppose I have two apps, and I use app 1 exclusively for data collection, but I used app 2 both to post tweets to my timeline interactively and occasionally in an automated way. Then I would conceivably have three different auth objects, which I would probably want to name: `app1-app`, `app2-user` and `app2-bot`. And I would want to be able to select amongst these.

hadley:
You'd switch between them using `auth_as("app1-app")` etc.

R/auth.R
```diff
@@ -0,0 +1,292 @@
...
+#' @export
+#' @rdname rtweet_user
+rtweet_bot <- function(
+      api_key = askpass::askpass("Please enter your API key"),
+      api_secret = askpass::askpass("Please enter your API secret"), 
+      access_token = askpass::askpass("Please enter your access token"), 
+      access_secret = askpass::askpass("Please enter your access token")
+  ) {
+  stopifnot(is_string(api_key), is_string(api_secret))
+  stopifnot(is_string(access_token), is_string(access_secret))
```
llrs, line 70: Provide a more informative error like missing more api_key or api_secret or...?

R/auth.R
```diff
@@ -0,0 +1,292 @@
...
+    path <- auth_path(paste0(auth, ".rds"))
+    if (!file.exists(path)) {
+      abort(paste0("Can't find saved auth with name '", auth, "'"))
+    }
+    inform(paste0("Reading auth from '", path, "'"))
+    readRDS(path)
+  } else {
+    abort("Unrecognised input to `auth`")
+  }
+}
```
llrs, line 212: I find it hard to follow, why there is a specific branch for dev_mode?
On my .Rprofile I set to load devtools, so it always detect that I'm on dev_mode and returns the rtweet_test(). Not sure how widespread this is but I expect all potential contributors and developers of dependencies to have loaded devtools. Perhaps it should test if it is on dev_mode by looking at the dev_path option set up by `devtools::dev_mode()`.

README.Rmd
```diff
@@ -110,22 +110,14 @@ All you need is a **Twitter account** (user name and password) and you can be up
...
-This uses the Twitter developer app baked into the package. If you want to post tweets, use a bearer token, or read/write direct messages, or work non-interactively, you will need to create your own app. See the details in `vignette("auth")`.
+Using your personal account is fine for casual use, but if you are trying to collect a lot of data it's a good idea to create your own twitter "app" and use `auth_app()`. Learn more in `vignette("auth")`.
 
 ### Vignettes
 
-[Obtaining and using Twitter API tokens](https://docs.ropensci.org/rtweet/articles/auth.html)
-
-```{r}
-## quick overview of rtweet functions
-vignette("auth", package = "rtweet")
```
llrs, line 125: I think this package = "rtweet" argument was added because there are other package with a vignette named auth (there was a PR or an issue opened by a user). Might be better to keep it and use it on all the instances we want to redirect the users to this vignette.

NEWS.md
```diff
@@ -1,5 +1,14 @@
 # rtweet ...
 
+- `create_token()` has been deprecated in favour of the combination of
+  `rtweet_user()`/`rtweet_bot()`/`rtweet_app()` + `auth_use()` + `auth_save()`.
+
+- `get_token()` and `get_tokens()` have been deprecated in favour of 
+  `auth_get()`.
```
llrs, line 7: `auth_get` is internal. I think this should be `auth_as`

llrs (changes requested):
Overall it is much easier now to know what is happening and the different use cases. Haven't had time to fully test it (or look into tests), but there are many comments.

vignettes/auth.Rmd
```diff
@@ -12,104 +12,76 @@ vignette: >
...
-`create_token()` will automatically cache the token for you, so the next time you start an R session on the same machine, rtweet should automatically find your token. To double check it works, restart your R session, run the following code, and check that the app name and `api_key` match.
+rtweet's default authentication mechanism allows you to act on behalf of your personal Twitter account, as if you were performing actions on twitter.com.
+This vignette shows you how to use two other authentication mechanisms useful for other scenarios:
+
+-   App authentication allows you to act as if you were a Twitter app.
+    You can't perform operations that a user would (like posting a tweet or reading a DM), but in return you get higher rate limits on operations like searching for tweets.
+
+-   Bot authentication allows you to create a fully automated Twitter bot that performs actions on its own behalf rather than on behalf of a human.
+
+In either case, you'll need to start by creating a Twitter app, so this vignette will first walk you through that process.
```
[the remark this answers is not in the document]
hadley: I think there's two things missing here:

* An explanation of what a twitter app is
* Mentioning that rtweet comes with a special one built-in

vignettes/auth.Rmd
```diff
@@ -12,104 +12,76 @@ vignette: >
...
+## Setup
+
+Now you're ready to tell rtweet how to authenticate using your app.
+You have two options:
+
+-   `rtweet_user()` is best for collecting data.
+-   `rtweet_bot()` is best for creating twitter bots.
+
+In both cases, you first create an auth object using data from the "tokens & keys" page then save it for use in future sessions.
+This process is a little trickier than you might expect because you never want to enter API keys into the R console or save in an .R script to avoid accidentally exposing secrets.
```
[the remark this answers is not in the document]
hadley: Hmmm, that uses the console but takes special case on your behalf. I'll see if I can word this better.

R/auth.R
```diff
@@ -0,0 +1,292 @@
...
+#' auth_as()
+#' 
+#' # Load auth saved by auth_save()
+#' auth_as("my-saved-app")
+#' }
+auth_as <- function(auth = NULL) {
+  old <- .state$auth
+  .state$auth <- find_auth(auth)
+  invisible(old)
+}
```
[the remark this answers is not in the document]
hadley: This is a standard pattern — whenever you have a function that affects some global state (e.g. `options()`, `par()`) it should always return the *previous* value. There's a little bit about why at the bottom of https://design.tidyverse.org/out-invisible.html

R/auth.R
```diff
@@ -0,0 +1,292 @@
...
+#' to set up.
+#' 
+#' @keywords internal
+#' @export
+auth_get <- function() {
+  if (is.null(.state$auth)) {
+    auth_as()
+  }
+  .state$auth
+}
```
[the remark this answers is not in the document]
hadley: I think a function that lists all available options should have a different name, like `auth_list()`.

R/auth.R
```diff
@@ -0,0 +1,292 @@
...
+#' # later, in a different session...
+#' auth_us("my-app")
+#' }
+auth_save <- function(auth, name) {
+  stopifnot(is_auth(auth), is_string(name))
+  
+  path <- auth_path(paste0(name, ".rds"))
+  inform(paste0("Saving auth to '", path, "'"))
+  
+  dir.create(auth_path(), showWarnings = FALSE, recursive = TRUE)
```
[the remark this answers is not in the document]
hadley: I don't think there's much point in telling a user that something has happened that they can't undo, and in general, functions that write to disk (e.g. `write.csv()`, `writeLines()`) do not warn that a file already exists.

R/auth.R
```diff
@@ -0,0 +1,292 @@
...
+  if (length(options) == 0) {
+    if (is_interactive()) {
+      inform("No saved authentication found; creating...")
+      auth_save(rtweet_user(), "default")
+    } else {
+      abort("No saved auth found; please resolve in an interactive session")
+    }
+  } else if (length(options) == 1) {
+    readRDS(options)
+  } else {
```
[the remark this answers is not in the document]
hadley: I'm fine with making that explicit; I only did that because I thought that's what you wanted.

R/auth.R
```diff
@@ -0,0 +1,292 @@
...
+#' @export
+rtweet_user <- function(api_key = NULL, api_secret = NULL) {
+  if (is.null(api_key) && is.null(api_secret)) {
+    decrypt <- function(x) {
+      rawToChar(openssl::rsa_decrypt(x[[2]], x[[1]]))
+    }
+    api_key <- decrypt(sysdat$DYKcJfBkgMnGveI)
+    api_secret <- decrypt(sysdat$MRsnZtaKXqGYHju)
+  } else {
+    stopifnot(is_string(api_key), is_string(api_secret))
```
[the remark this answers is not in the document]
hadley: I don't think it's worth investing time in error messages that should only be seen by developers (and who have supplied fundamentally the wrong type of input to a function).

R/auth.R
hadley: Oh you're worried about the case where the user doesn't provide an info to `askpass()`? That would be better handled with a wrapper function.

R/auth.R
hadley: This only applies if you have loaded **rtweet** with devtools, and thus prevents you running the tests with your personal account.

NEWS.md
hadley: `auth_get()` retrieves the current auth; it's not something that most users will need.

vignettes/auth.Rmd
```diff
@@ -12,104 +12,76 @@ vignette: >
...
+auth <- rtweet_app()
+auth <- rtweet_bot()
+```
+
+You will then get a series of pop-ups asking you to enter data from the "keys & tokens" page:
+
+-   `rtweet_app()` is easy: you need to enter the bearer token, which you can get by clicking the "regenerate" button next to "bearer token".
+
+-   `rtweet_bot()` is a bit more work because you need four pieces of data: the API key and secret, and the access token and secret.
+    To get these two pairs of data, you'll need to click the two regenerate buttons.
```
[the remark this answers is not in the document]
hadley: I could've sworn I didn't see that when I created an app yesterday, but I do now. I'll update and include a screenshot.

R/auth.R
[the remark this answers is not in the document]
hadley: How about we have an explicit `auth_setup()` that does `auth_save(auth_user(), "default")`. Then this function would use `default.rds` if available, otherwise it would error with informative instructions.  

This means that you'd need to call `auth_setup()` once the first time you use rtweet 1.0.0.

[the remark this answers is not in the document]
llrs: mmh, yes I want the default to be saved but knowing what I'm saving with an informative name, so that when using auth_as next time I know which token I'm using. Default is not enough informative because there are two auth defaults provided by rtweet: rtweet_app and rtweet_user.

R/auth.R
llrs: I am worried that the error won't be informative enough for some users. Not sure what happens when askpass is left empty.

R/auth.R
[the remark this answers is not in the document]
llrs: I'm ok with the new name although I would prefer to keep it. But rtweet should not only give feedback on existing options on an error message (when auth_as finds multiple tokens saved).

R/auth.R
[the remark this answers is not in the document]
hadley: I'm not sure what you mean here — there is only one default auth provided by rtweet: `rtweet_user()`.

hadley:
Ok, I've taken another pass at it — `auth.Rmd` hopefully now emphasises that the name is arbitrary, and it just has to be meaningful to you. I've also switched the default auth to require an initial explicit call to `auth_setup_default()`, and it only ever looks for `default.rds` (

vignettes/auth.Rmd
```diff
@@ -12,104 +12,100 @@ vignette: >
...
+It's a good idea to do this once to check that you've entered everything correctly, but it'd be annoying if you had to pass this object around every single time.
+Instead, you can use `auth_use()` to set this as the default for the remainder of the session:
+
+```{r, eval = FALSE}
+auth_as(auth)
+```
+
+## Saving and loading
+
+`auth_use()` only lasts for a single session; if you close and re-open R, you'll need to repeat the whole process.
```
alexpghayes, line 97:
```suggestion
`auth_as()` only lasts for a single session; if you close and re-open R, you'll need to repeat the whole process.
```

Typo?

hadley:
I've taken a couple more passes through the docs

NEWS.md
```diff
@@ -26,21 +26,33 @@
...
   endpoint in June 2019 (https://twittercommunity.com/t/124732).
 
-- rtweet no longer touches your `~/.Renviron`. Instead, it now caches the last
-  used token in a system specific cache directory.
-
 - Added support for posting alt-text metadata with images tweeted with status updated via `post_tweet()`. (@hrbrmstr)
 
+## Authentication
+
+rtweet's authentication system has been completely written. It is now based around three authentication options: `rtweet_user()`, `rtweet_app()`, and `rtweet_bot()`. Authenetication no longer touches `~/.Renviron` file; instead `auth_save()` and `auth_as()` allows you to explicitly save an load authentication mechanisms from a system config directory. See `vignette("auth")` for more details.
```
alexpghayes, line 40:
```suggestion
rtweet's authentication system has been completely written. It is now based around three authentication options: `rtweet_user()`, `rtweet_app()`, and `rtweet_bot()`. Authentication no longer touches `~/.Renviron` file; instead `auth_save()` and `auth_as()` allow you to explicitly save and load authentication mechanisms from a system config directory. See `vignette("auth")` for more details.
```

NEWS.md
```diff
@@ -26,21 +26,33 @@
...
+
+- The httpuv package is now only suggested, since it's only needed for 
+  interactive auth, and you'll be prompted to install it when needed.
+
+- `bearer_token()` has been deprecated in favour of `rtweet_app()`, which takes 
+  the bearer token found in your Twitter developer portal. `invalidate_bearer()`
+  has been deprecated since this is something you should do yourself in the
+  Twitter developer portal.
+
+- `create_token()` has been deprecated in favour of the combination of
```
alexpghayes, line 50: This line really helped things click for some reason, thanks!

llrs (approved):
This looks great! There are several function that do not have the `#' @return` tag, it would be nice if all of the functions would have it, but this can be added later. 

Also the README hasn't been build with the latest changes, but as I'm sure it  will change soon, it might not worth your time to spend on it.

PR #543, opened 2021-03-24
Standardize DESCRIPTION URLs

jennybc:
This is more typical of what we do in tidyverse / r-lib and, as a result, works better with usethis's `browse_*()` functions.


jennybc:
I don't think the test failures have anything to do with my URL change! But maybe they are interesting anyway? There seems to be at least 2 different things going on.

hadley:
The auth failures are because we accidentally got the test user blocked 😬

PR #547, opened 2021-03-30
Revise readme

### LINKED ISSUE #544: Move feature comparison further down the home page
I think it would make sense to move this table either further down the page or to a separate vignette (or removed all together). None of the other packages have been updated for years and rtweet has so many more features, it starts to feel a little mean.

hadley:
* Light pruning of badges
* Remove outdated comparison table
* Simplify install instructions
* Focus on in key use cases
* Remove figures since they were broken on the website, bloat the package size, and involve functions we no longer want to promote
* Rename intro.Rmd to rtweet.Rmd
* Revise package docs

Fixes #544


llrs (approved):
Not sure why the status badges were deleted (both lifecycle and repostatus), but the rOpenSci badge must remain.

Also there's a double curly bracket on rtweet: change {{rtweet}} to {rtweet}?

hadley:
I don't have strong feelings about any of the badges; feel free to add/remove as you see fit.

PR #551, opened 2021-04-01
Make more examples live

hadley:
These will not run on CRAN, but they will run in our GitHub actions, and for authenticated users who call `example()`. Since they're now live, I've generally trimmed the examples to take less time. Since I was touching the code I also generally replaced individual political figures with pop stars.


llrs:
I have recently incorporated this. Thanks!

PR #554, opened 2021-04-02
Standardise documentation for pagination parameters

hadley:
I've tried to document all parameters related to pagination in one place: the documentation for `TWIT_paginate_max_id`/`TWIT_paginate_cursor`. This means that the docs lose less specific details about the endpoint (e.g. specific page sizes and rate limits) but gain more general details about how all the pieces fit together.

@llrs I think the best place to focus on is the `TWIT_paginate_max_id` documentation, since that now flows out to the majority of functions in rtweet. Let me know if there's anything you think is confusing so that I can improve the wording as much as possible.

### DIFF
--- a/R/http.R
+++ b/R/http.R
@@ -61,14 +61,26 @@ TWIT_method <- function(method, token, api,
 
 #' Pagination
 #' 
+#' @description 
+#' `r lifecycle::badge("experimental")`
+#' These are internal functions used for pagination inside of rtweet.
+#' 
 #' @keywords internal
+#' @param token Expert use only. Use this to override authentication for
+#'   a single API call. In most cases you are better off changing the
+#'   default for all calls. See [auth_as()] for details.
 #' @param n Maximum number of results to return.
 #' @param get_id A single argument function that returns a vector of ids given 
 #'   the JSON response. The defaults are chosen to cover the most common cases,
 #'   but you'll need to double check whenever implementing pagination for
 #'   a new endpoint.
-#' @param max_id String giving id of most recent tweet to return. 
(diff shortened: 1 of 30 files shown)


R/http.R
```diff
@@ -61,14 +61,38 @@ TWIT_method <- function(method, token, api,
...
 #'   a new endpoint.
-#' @param max_id String giving id of most recent tweet to return. 
-#'   Can be used for manual pagination.
+#' @param max_id Provide an upper bound for the returned results; all results
+#'   will have an ID less than or equal to `max_id`. If omitted, results will 
+#'   start from the most recent available. Supplying `max_id` allows you to 
+#'   manually paginate the results.
+#' @param since_id Provides a lower-bound for the results returned; all results
+#'   will have an ID greater than or equal to `since_id`. This can generally
+#'   be omitted.
```
llrs, line 95: Should it be a numeric ID or a character? Not sure I understand well what this is but each message or status has an ID this ID increases as more content is published so if one wants content later than X message can use this?

llrs (approved):

R/http.R
hadley: Yes, that's right. But once #510 is done, the user shouldn't need to worry about the details. e.g. the code for retrieving more tweets from a search would be something like this:

```R
tw <- search_tweets("#rstats", n = 10000)
tw_more <- search_tweets("#rstats", max_id = max_id(tw))
```

And that makes me realise why you might need `since_id()` — this code would get all the tweets since your last request:

```R
tw_more <- search_tweets("#rstats", max_id = since_id(tw))
```

I'll make all this work in upcoming PRs.

PR #556, opened 2021-04-02
Consistent interface for paginating tweets

hadley:
i.e. all functions that use `TWIT_paginate_max_id()`.

This PR restores `max_id` and `since_id` to life, and uses them automatically in `TWIT_paginate_max_id()` so that you can pass in previously retrieved data.


hadley:
@llrs two changes here, one smaller, one bigger. Now all functions that use `TWIT_paginate_max_id()` have the `max_id`, `since_id`, `retryonpagelimit` (I ran out of energy to come up with a better name), and `verobse` arguments. I also retrofitted the `max_id()` and `since_id()` arguments so that they could be used by `TWIT_paginate_max_id()`; that makes it very easy to retrieve results either older or newer than what you already have.

tests/testthat/test-next_cursor.R
```diff
@@ -0,0 +1,22 @@
+test_that("max_id and since_id work in bit64", {
+  expect_equal(max_id("0"), "-1")
+  expect_equal(max_id(c("50", "100")), "49")
+  
+  expect_equal(since_id(c("50", "100")), "100")
+})
```
llrs, line 6: Isn't the name a bit misleading? Wouldn't a better test name be max_id and since_id work in characters? Maybe there should be a different test for bit64 numbers.

R/favorites.R
```diff
@@ -1,22 +1,21 @@
...
 #' @inheritParams TWIT_paginate_max_id
 #' @inheritParams get_timeline
 #' @return A tibble with one row for each tweet.
 #' @examples
 #' \dontrun{
-#'
-#' ## get max number of statuses favorited by KFC
-#' kfc <- get_favorites("KFC", n = 3000)
+#' # get likes for a single user
+#' kfc <- get_favorites("KFC", n = 600)
```
llrs, line 11: Perhaps make the examples faster by reducing to n = 200 to make just one request?

R/http.R
```diff
@@ -119,6 +116,12 @@ TWIT_paginate_max_id <- function(token, api, params,
                                  count_param = "count", 
                                  retryonratelimit = FALSE,
                                  verbose = TRUE) {
+  if (!is.null(max_id)) {
+    max_id <- rtweet::max_id(max_id)  
+  }
+  if (!is.null(since_id)) {
+    since_id <- rtweet::since_id(since_id)
```
llrs, line 123: Why the rtweet::max_id and rtweet::since_id? Isn't max_id and since_id enough?

R/http.R
```diff
@@ -161,7 +164,7 @@ TWIT_paginate_max_id <- function(token, api, params,
       break
     }
     
-    max_id <- id_minus_one(last(get_id(json)))
+    max_id <- max_id(get_id(json))
```
llrs, line 167: I have missed this but json is used on the `is_rate_limit(resp)` before it is generate on `json <- from_js(resp)` (Thanks Rstudio for the static code analysis :). These lines will probably need to be reordered.

R/favorites.R
```diff
@@ -1,22 +1,21 @@
...
 #' kfc
+#' # get newer likes since last request
+#' newer <- get_favorites("KFC", since_id = kfc)
```
llrs, line 14: Perhaps restrict kfc to some rows to show some results otherwise unless they tweet between these two calls newer will be always empty (this also happens on other examples)

llrs (approved):
I think this will be a nice addition for those trying to collect data periodically. I've made some suggestions about a examples, some test and internal code.

tests/testthat/test-next_cursor.R
hadley: They work _in_ bit64 space with character vectors.

R/favorites.R
hadley: I already made it five times faster 😛 But yes, smaller is better. There's probably no reason to ever set `n` in examples, unless we specifically want to show pagination.

R/http.R
hadley: I think it makes the code a little easier to follow than `since_id <- since_id(since_id)`

R/http.R
hadley: Ooops, good catch!

R/favorites.R
hadley: Ok, I'll remove it from the example here and instead put it in `search_tweets` where it is likely you'll see new data.

PR #560, opened 2021-04-06
Consistent interface for cursor based pagination

hadley:
The only function here where paginations seems likely to be used is `get_followers()` so I concentrated my testing on it.

Part of #510


llrs (approved):
I was surprised by copy_cursor I expected to be copy_cursor(from, to). This change would basically to swap all the arguments where the functions appear.

hadley:
I started with `copy_cursor()` like that but I think mildly prefer this arrangement better because it's the first argument that's returned (modified by the second).

PR #562, opened 2021-04-07
Take default value of retryonratelimit from an option

### LINKED ISSUE #173: Feature request: global retryonratelimit
`python-twitter` has this very neat feature: when you instantiate the `Api` object, there's a `sleep_on_rate_limit` parameter, so that all the methods have this either enabled or disabled (see [here](http://python-twitter.readthedocs.io/en/latest/rate_limits.html)). I think that it would be a nice addition for `rtweet` too.

hadley:
So you can easily set it to TRUE everywhere, if desired. Fixes #173.

@llrs this is probably my last PR for a while; I need to get back to other stuff, and I think rtweet is now in pretty solid shape.


llrs (approved):
Many thanks @hadley for all the contributions on rtweet. It is indeed in much better in my opinion. Hope to keep it in good shape for future developments.

PR #564, opened 2021-04-09
return parameters from my_list_params

noamross:
`my_list_params()` was not returning the parameters object properly to pass on to the `post_list*` functions.  This fixes that.


llrs:
Many thanks

Issue #566, opened 2021-04-12
get_favorites n=Inf error


jsta:
The docs for `get_favorites` suggest:

> Set n = Inf to download as many results as possible

However, when I do this I get the following error:

> Error in vector("list", pages) : vector size cannot be infinite

llrs:
Thanks for the report. I'll look into it.

PR #567, opened 2021-04-15
Prevent infinite requests

llrs:
This is to protect the user from infinite requests on #566 by setting a new value the max number of requests possible on a 15 minutes period. Fixed the same for some the other functions that use n, now they should at most hit the rate limit and finish then. 

If a users decided to explore the limits of the package it will save the limits of the auth method and store it on the first call for later reuse.

Found some trouble with search_users. I think it will need a new paginate function. It uses the TWIT_paginate_max_id documentation but neither of the three functions are called. It cannot use the cursor and not sure how to solve it yet.
TODO: update the TWIT_paginate_max_id documentation of n to explain the new behaviour of n.


hadley:
This doesn't look like the right approach to me — since it mostly affects `TWIT_paginate_max_id()`, I'd think you'd want to mostly handle it there. If `n` is `Inf` you'd want to start `results` with some fairly large size, and then in the loop check to see if you're about to run out of result space and if so, double the size of the list.

llrs:
I started thinking it was an easy fix then I realized I was doing something much more complicated and I was not longer sure if I was taking the right approach. 

I'm don't understand what you mean with "result space". If the user says n = Inf and then I start with a vector of size 99999, you propose to extend the vector to size 99999*2 when it reaches 99999 results. Either the rate limit will be hit and the function will halt or if `rateonlimit = TRUE` it will continue to extend the vector until their computer cannot longer work with such a list (at which point all the results will be lost). 

I think it would be easier for users to prevent this situation by setting a big number that wouldn't crash users' computer and let them continue using the appropriate method to continue later (max_id, since_id, ...).

hadley:
You start the vector with (say) 1000 elements. If you get to 1001 elements and still aren't finished, you grow the vector to 2000 elements. On average that takes roughly the same amount of time as setting the vector to the correct size (if you new what it was), without having to assign a massive list to begin with. 

I don't think the primary concern should be running out of memory, unless that's something you've seen in practice?

llrs:
Initially I thought it would be an easy fix but end up doing this complicated PR. I haven't seen a case of running out of memory using rtweet. 

Anyway I tried your suggestion but then I got blocked by how to correctly report the progress to the user. Opened a question on [Community Rstudio](https://community.rstudio.com/t/progress-update-bar/102160) about the progress bar. Anyway, closing the issue as I won't continue with this PR

Issue #558, opened 2021-04-05
Reduce tweet fields returned by default


hadley:
Current `search_tweets()` and friends returns a data frame with 73 columns:

```R
 [1] "status_id"               "created_at"              "user_id"                
 [4] "screen_name"             "text"                    "source"                 
 [7] "display_text_width"      "reply_to_status_id"      "reply_to_user_id"       
[10] "reply_to_screen_name"    "is_quote"                "is_retweet"             
[13] "favorite_count"          "retweet_count"           "quote_count"            
[16] "reply_count"             "hashtags"                "symbols"                
[19] "urls_url"                "urls_t.co"               "urls_expanded_url"      
[22] "media_url"               "media_t.co"              "media_expanded_url"     
[25] "media_type"              "ext_media_url"           "ext_media_t.co"         
[28] "ext_media_expanded_url"  "ext_media_type"          "ext_alt_text"           
[31] "mentions_user_id"        "mentions_screen_name"    "lang"                   
[34] "quoted_status_id"        "quoted_text"             "quoted_created_at"      
[37] "quoted_source"           "quoted_favorite_count"   "quoted_retweet_count"   
[40] "quoted_user_id"          "quoted_screen_name"      "quoted_name"            
[43] "quoted_followers_count"  "quoted_friends_count"    "quoted_statuses_count"  
[46] "quoted_location"         "quoted_description"      "quoted_verified"        
[49] "retweet_status_id"       "retweet_text"            "retweet_created_at"     
[52] "retweet_source"          "retweet_favorite_count"  "retweet_retweet_count"  
[55] "retweet_user_id"         "retweet_screen_name"     "retweet_name"           
[58] "retweet_followers_count" "retweet_friends_count"   "retweet_statuses_count" 
[61] "retweet_location"        "retweet_description"     "retweet_verified"       
[64] "place_url"               "place_name"              "place_full_name"        
[67] "place_type"              "country"                 "country_code"           
[70] "geo_coords"              "coords_coords"           "bbox_coords"            
[73] "status_url"  
```

I'd suggest that we return fewer more complicated columns by default, instead providing some helpers to access them when needed. For example, we could keep `media_*`, `quoted_*` and `retweet_*` in `media`, `quoted` and `rtweet` columns and provide helpers to expand them out when needed.

hadley:
IOTW I'm suggesting that the data frame unpacking that currently occurs in `tweets_to_tbl_()` should be post-poned until the user requests it.

[...]

PR #572, opened 2021-04-22
Simplify response parsing

llrs:
Simplify object parsing towards #558 

Steps:
- [x] Internal test for [objects](https://developer.twitter.com/en/docs/twitter-api/v1/data-dictionary/object-model/user)
- [x] Simplify and understand code
- [x] Provide helpers

----

`tweets_to_tbl_` is doing the internal tests and transforming the data.frame from X columns to Y columns for all these functions:

function/endpoint | From | To
------|----|---
get_retweets | 24 | 73
get_timeline_user |  |
lookup_tweets | | 
lookup_users |  |
get_favorites_user |  |
get_timeline |  |
lists_status |  |
get_my_timeline |  |
search_users| | 
search_tweets|  |
post_message |  | 
search_fullarchive |  |
search_30day |  |
tweet_shot |  |

*Numbers from the test cases. 

Probably a list of endpoints that supply each kind of object will be much useful.


llrs:
Updating the table:

function/endpoint | From | To
------|----|---
get_retweets | 24 | 73
get_timeline_user | 31 | 73
lookup_tweets | 26| 73
lookup_users | 44 | 20
get_favorites_user | 30 | 73 + 1 
get_timeline | 31 | 73
lists_status | 31 | 73
get_my_timeline | 32 | 73
search_users| 44 | 20
search_tweets| 31 | 73

On some cases there is less columns when the content is parsed than when not (lookup_users, search_users)! This is due to the `users_with_tweets` function, but apparently the API is returning lots of data from deprecated arguments but misses some data from supported values. 
I'll start improving the user-object

The 73 + 1  is the user added when parsed to know which user has liked which tweet.

llrs:
Documenting progress, about blocking problems at the moment:
- Tweets, and all responses,  are now parsed based on the objects they contain. 
- The latest changes trigger a warning: "Setting row names on a tibble is deprecated.", although row names are not used in any place to parse tweets. It only happens on `do.call("rbind", )` from `tweets_with_users` and `users_with_tweets`. 
- For instance `out <- get_favorites(users, n = 20)` triggers an error `z[seq_len(nro), ] <- y : replacement has length zero` on `xpdrows.data.frame`. Which is surprising and couldn't understand why or how it is triggered. I think it is related to having data.frames inside data.frames inside data.frames. 
- The new data.frame has much less columns and different names to keep consistent with the API, so functions (and the corresponding test) that depend on certain columns fail. 
- Due to the above problems, it hasn't been tested if the tweets object has the same columns regardless of the endpoint called (Is this desired?) .

llrs:
Many changes, but still need to add a test for #574 (and didn't check for #575 but should be fixed too)

- Added some test for different type of tweets so that rtweet returns the same information
- The columns are on the same order
- Localization and other objects are now correctly identified/used
- Some reorganizing on network_data

Maybe some columns can be further nested or hidden on an attribute.
There isn't any helper to identify replies, comments, quotes and simply status. 
Moving away of tibbles as nested tibbles cannot be rbinded without a warning and would mean adding a new direct dependency https://github.com/tidyverse/tibble/issues/898

llrs:
@hadley Maybe you can give a look at this PR. Probably the best way to start is on the tweet function and the new functions to parse twitter objects.

R/entities_objects.R
```diff
@@ -0,0 +1,114 @@
+# Contains functions to parse the objects described here: 
+# https://developer.twitter.com/en/docs/twitter-api/v1/data-dictionary/object-model/entities
+
+# <https://developer.twitter.com/en/docs/twitter-api/v1/data-dictionary/object-model/entities#hashtags>
+hashtags <- function(x) {
+  if (NROW(x) == 0) {
```
hadley, line 6: The use of `NROW()` suggests the input is going to be either a vector or a matrix.

R/entities_objects.R
```diff
@@ -0,0 +1,114 @@
...
+    return(data.frame(text = NA, indices = I(list(NA)), 
+                      stringsAsFactors = FALSE))
+  }
```
hadley, line 9: I think this function would be clearer if written more symmetrically:

```R
if (nrow(x) == 0) {
   data.frame(...)
} else {
  i <- indices_vec(x$indices)
  data.frame(...)
}
```

R/entities_objects.R
```diff
@@ -0,0 +1,114 @@
...
+  i <- indices_vec(x$indices)
+  data.frame(text = x$text, indices = I(i))
+}
+
+
+indices_vec <- function(x) {
+  i <- lapply(x, function(y){matrix(y, ncol = 2, 
+                                    dimnames = list(NULL, 
+                                                    c("start", "end")))})
```
hadley, line 18: Indenting makes this a little hard to read, and the assignment is unnecessary.

R/entities_objects.R
```diff
@@ -0,0 +1,114 @@
...
+  i
+}
+
+# The extended entities is really for media
+# <https://developer.twitter.com/en/docs/twitter-api/v1/data-dictionary/object-model/extended-entities>
+media <- function(x) {
+  df <- data.frame("id" = NA, "id_str" = NA, "indices" = I(list(NA)),
```
hadley, line 25: Don't need quotes around names?

R/entities_objects.R
```diff
@@ -0,0 +1,114 @@
+# Contains functions to parse the objects described here: 
+# https://developer.twitter.com/en/docs/twitter-api/v1/data-dictionary/object-model/entities
+
+# <https://developer.twitter.com/en/docs/twitter-api/v1/data-dictionary/object-model/entities#hashtags>
+hashtags <- function(x) {
+  if (NROW(x) == 0) {
+    return(data.frame(text = NA, indices = I(list(NA)),
```
hadley, line 7: Is it correct that a zero row input yields a one row output?

R/entities_objects.R
```diff
@@ -0,0 +1,114 @@
...
+  x[setdiff(colnames(df), colnames(x))] <- rep(NA, nrow(x))
+  rownames(x) <- NULL
+  x
+}
+
+# PowerTrack $ text
+# has:symbol
+# They are the same
+# <https://developer.twitter.com/en/docs/twitter-api/v1/data-dictionary/object-model/entities#symbols>
+symbols <- hashtags
```
hadley, line 78: Would be a bit simpler to understand if you just had one function.

R/favorites.R
```diff
@@ -45,9 +45,10 @@ get_favorites <- function(user,
 }
 
 get_favorites_user <- function(user, ..., parse = TRUE, token = NULL) {
+  stopifnot(length(user) == 1)
   params <- list(
-    tweet_mode = "extended",
-    include_ext_alt_text = "true"
+    # Undocumented parameter https://github.com/ropensci/rtweet/issues/575#issuecomment-829605892
```
hadley, line 50: Nice pointer 😄

R/geo_objects.R
```diff
@@ -0,0 +1,56 @@
+# Contains functions to parse the objects described here: 
+# https://developer.twitter.com/en/docs/twitter-api/v1/data-dictionary/object-model/geo
+
+bounding_box <- function(x) {
+  empty <- data.frame(long = NA, lat = NA, type = NA)
+  if(is.null(x) || length(x) == 1 && is.na(x)) {
```
hadley, line 6:
```suggestion
  if (is.null(x) || (length(x) == 1 && is.na(x))) {
```

hadley: Another check would be `is.null(x) || identical(x, NA)` (assuming that it's always a logical NA)

R/geo_objects.R
```diff
@@ -0,0 +1,56 @@
...
+    coord <- x$coordinates[[1]][1, , ]
+    if (is.null(coord)) {
+      return(empty)
+    }
+    return(data.frame(long = coord[, 1], lat = coord[, 2], type = I(x$type)))
+  }
+  m <- x$coordinates[1, , ]
+  colnames(m) <- c("long", "lat")
+  df <- as.data.frame(m)
+  df$type <- I(x$type)
```
hadley, line 19: You only need to use `I()` when you're creating a list-column with `data.frame()`.

R/graph-network.R
```diff
@@ -102,75 +42,144 @@ prep_from_to <- function(x, from, to) {
 #'   }
 #' }
 #' @export
-network_data <- function(.x, .e = c("mention,retweet,reply,quote")) {
-  if (isTRUE(.e)) {
-    .e <- "all"
+network_data <- function(x, e = c("mention", "retweet","reply", "quote")) {
+  if (isTRUE(e) || length(e) == 1 && (e == "semantics" || e == "all")) {
```
hadley, line 46:
```suggestion
  if (isTRUE(e) || (length(e) == 1 && e %in% c("semantics", "all"))) {
```

R/graph-network.R
```diff
@@ -102,75 +42,144 @@ prep_from_to <- function(x, from, to) {
...
+    e <- c("mention", "retweet","reply", "quote")
```
hadley, line 47:
```suggestion
    e <- c("mention", "retweet", "reply", "quote")
```

R/utils.R
```diff
@@ -100,7 +104,7 @@ maybe_n <- function(x) {
 }
 
 is_testing <- function() {
-  identical(Sys.getenv("TESTTHAT"), "true")  
+  requireNamespace("testthat", quietly = TRUE) && identical(Sys.getenv("TESTTHAT"), "true")
```
hadley, line 107: That check shouldn't be needed since only testthat should be setting the envvar.

tests/testthat/test-entities_objects.R
```diff
@@ -0,0 +1,210 @@
+hashtags <- jsonlite::fromJSON('{
```
hadley, line 1: I'd recommend putting the test data inside the tests to keep each test self contained.

tests/testthat/test-extractors.R
```diff
@@ -3,12 +3,12 @@ test_that("tweets_data works", {
 
   ## get data on most recent tweet from user(s)
   tweets <- tweets_data(jack)
-  expect_s3_class(tweets, "tbl_df")
-  expect_true("status_id" %in% names(tweets))
+  expect_s3_class(tweets, "data.frame")
```
hadley, line 6: Is it a deliberate change to go from tibbles to data.frames? data.frames are not great at printing list-columns.

hadley (commented):
Looks like some great improvements overall! Questions/comments below.

llrs:
Many thanks for all the comments. Will try to reply and address them soon (1 or 2 weeks probably).

R/entities_objects.R
llrs: Yes, I use it in case the object is a NULL vector.

R/entities_objects.R
llrs: Done!

R/entities_objects.R
llrs: Hope now it is more readable

R/entities_objects.R
llrs: Correct, removed the quotes from other data.frames too

R/entities_objects.R
llrs: Yes, the reason is to not omit anything for later easier parsing and linking between hasthags and tweets. There is also the problem when a user hasn't tweeted if the information about the tweets of that users and other users are collected it would break the tweet_data extraction see #574 for such a report.

R/entities_objects.R
llrs: On `parse_entities2` I use the name of the columns to match with the appropriate function to parse it. So I needed a symbols function and it matched with the same of what hasthags function was doing so I created it that way.

R/geo_objects.R
llrs: I corrected the style and changed to use the identical check. It didn't break any test, so I think it is safe to assume it will be always a logical NA.

R/geo_objects.R
llrs: Somehow I thought it was a data.frame. Removed.

R/graph-network.R
llrs: Thanks, it is more readable

R/utils.R
llrs: It should because later on there are some function calling testthat function inside the package (skip_*). I swapped the order though so that if the variable is not present the package won't be loaded.

tests/testthat/test-entities_objects.R
llrs: Moved them in

tests/testthat/test-extractors.R
llrs: Yes, I was getting some warnings that I reported here: https://github.com/tidyverse/tibble/issues/898 and I decided to go back to data.frames to avoid users (and the package) getting that warning. Perhaps it should be transformed to a tibble before returning to the users via tweets_data...

R/entities_objects.R
hadley: In that case, I'd put this function right next to `hashtags` and mention why you define it (i.e. what you said above) in a comment.

tests/testthat/test-entities_objects.R
hadley: It looks like you still define the data outside of `test_that()`?

tests/testthat/test-extractors.R
hadley: Yeah. I think tibble should eliminate that warning — I'll raise this internally.

R/utils.R
hadley: I don't think it's a good idea to change `is_testing()` because it's a standard helper you'd generally expect to be copied directly from testthat. Where is `skip()` being used?

R/entities_objects.R
hadley: Right, I think you always want to return a data frame here, but are you sure it should have one row and not zero rows?

R/entities_objects.R
llrs: I tried to replace some by is.null but it didn't pass the checks then. So it is a combination of both.

R/entities_objects.R
llrs: Yes I'm sure. What is your concern? Filling the object with mostly empty data?

R/utils.R
llrs: There are two cases where this happens; on http.R lines 309-312:

```r
handle_rate_limit <- function(x, api, retryonratelimit = NULL, verbose = TRUE) {
  if (is_testing()) {
    testthat::skip("Rate limit exceeded")
  }
....
```
and auth.R lines 273-275 
```r
no_token <- function() {
  if (is_testing()) {
    testthat::skip("Auth not available")
...
```

I could move the check outside the function, but I thought it was more accurately to leave it inside it.

R/entities_objects.R
hadley: Just generally reasoning about this function. I don't understand how all the pieces fit together but this feels like a code smell.

llrs: I made some tests and it seems that mixing `NA`s and `data.frames` would be correctly handled by `rbind` unless there are nested `data.frame`s. I'll wait a bit to merge the PR but I can't think of any workaround that won't complicate more the workflow. 

<details>
<summary>Tests</summary>

``` r
``` r
l <- list(a = NA, b = data.frame(a = "b", b = "a"), c = data.frame(a = c("C", "D"), b = c("B", "C")))
do.call(rbind, l)
#>        a    b
#> a   <NA> <NA>
#> b      b    a
#> c.1    C    B
#> c.2    D    C
l <- list(a = data.frame(a = "b", b = "a"), b = NA, c = data.frame(a = c("C", "D"), b = c("B", "C")))
do.call(rbind, l)
#>        a    b
#> a      b    a
#> b   <NA> <NA>
#> c.1    C    B
#> c.2    D    C
l <- list(a = data.frame(a = "b", b = "a"), b = NA, c = data.frame(a = "C", b = "B"))
do.call(rbind, l)
#>      a    b
#> a    b    a
#> b <NA> <NA>
#> c    C    B
l <- list(a = data.frame(a = "b", b = "a"), b = data.frame(a = c("C", "D"), b = c("B", "C")), c = NA)
do.call(rbind, l)
#>        a    b
#> a      b    a
#> b.1    C    B
#> b.2    D    C
#> c   <NA> <NA>
library("tibble")
l <- list(a = NA, b = tibble(a = "b", b = "a"), c = tibble(a = c("C", "D"), b = c("B", "C")))
do.call(rbind, l)
#> # A tibble: 4 x 2
#>   a     b    
#> * <chr> <chr>
#> 1 <NA>  <NA> 
#> 2 b     a    
#> 3 C     B    
#> 4 D     C
l <- list(a = tibble(a = "b", b = "a"), b = NA, c = tibble(a = c("C", "D"), b = c("B", "C")))
do.call(rbind, l)
#> # A tibble: 4 x 2
#>   a     b    
#> * <chr> <chr>
#> 1 b     a    
#> 2 <NA>  <NA> 
#> 3 C     B    
#> 4 D     C
l <- list(a = tibble(a = "b", b = "a"), b = NA, c = tibble(a = "C", b = "B"))
do.call(rbind, l)
#> # A tibble: 3 x 2
#>   a     b    
#> * <chr> <chr>
#> 1 b     a    
#> 2 <NA>  <NA> 
#> 3 C     B
l <- list(a = tibble(a = "b", b = "a"), b = tibble(a = c("C", "D"), b = c("B", "C")), c = NA)
do.call(rbind, l)
#> # A tibble: 4 x 2
#>   a     b    
#> * <chr> <chr>
#> 1 b     a    
#> 2 C     B    
#> 3 D     C    
#> 4 <NA>  <NA>


l <- list(a = NA, b = data.frame(a = I(data.frame(c = 1)), b = "a"), c = data.frame(a = I(data.frame(c = c(2, 3))), b = c("B", "C")))
do.call(rbind, l)
#> Warning: non-unique value when setting 'row.names': '1'
#> Error in `.rowNamesDF<-`(x, value = value): duplicate 'row.names' are not allowed
l <- list(a = data.frame(a = I(data.frame(c = 1)), b = "a"), b = NA, c = data.frame(a = I(data.frame(c = c(2, 3))), b = c("B", "C")))
do.call(rbind, l)
#> Warning: non-unique values when setting 'row.names': '1', '2'
#> Error in `.rowNamesDF<-`(x, value = value): duplicate 'row.names' are not allowed
l <- list(a = data.frame(a = I(data.frame(c = 1)), b = "a"), b = NA, c = data.frame(a = I(data.frame(c = c(2))), b = "B"))
do.call(rbind, l)
#> Warning: non-unique value when setting 'row.names': '1'
#> Error in `.rowNamesDF<-`(x, value = value): duplicate 'row.names' are not allowed
l <- list(a = data.frame(a = I(data.frame(c = 1)), b = "a"), b = data.frame(a = I(data.frame(c = c(2, 3))), b = c("B", "C")), c = NA)
do.call(rbind, l)
#> Warning: non-unique value when setting 'row.names': '1'
#> Error in `.rowNamesDF<-`(x, value = value): duplicate 'row.names' are not allowed
l <- list(a = NA, b = tibble(a = I(data.frame(c = 1)), b = "a"), c = tibble(a = I(data.frame(c = c(2, 3))), b = c("B", "C")))
do.call(rbind, l)
#> Warning: non-unique value when setting 'row.names': '1'
#> Error in `.rowNamesDF<-`(x, value = value): duplicate 'row.names' are not allowed
l <- list(a = tibble(a = I(data.frame(c = 1)), b = "a"), b = NA, c = tibble(a = I(data.frame(c = c(2, 3))), b = c("B", "C")))
do.call(rbind, l)
#> Warning: non-unique values when setting 'row.names': '1', '2'
#> Error in `.rowNamesDF<-`(x, value = value): duplicate 'row.names' are not allowed
l <- list(a = tibble(a = I(data.frame(c = 1)), b = "a"), b = NA, c = tibble(a = I(data.frame(c = c(2))), b = "B"))
do.call(rbind, l)
#> Warning: non-unique value when setting 'row.names': '1'
#> Error in `.rowNamesDF<-`(x, value = value): duplicate 'row.names' are not allowed
l <- list(a = tibble(a = I(data.frame(c = 1)), b = "a"), b = tibble(a = I(data.frame(c = c(2, 3))), b = c("B", "C")), c = NA)
do.call(rbind, l)
#> Warning: non-unique value when setting 'row.names': '1'
#> Error in `.rowNamesDF<-`(x, value = value): duplicate 'row.names' are not allowed
```

<sup>Created on 2021-07-13 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

</details>

hadley: I'm probably missing something, but I don't see you creating an zero row data frames in your tests?  But I don't really understand how all of this data flows together, and there's likely to be something upstream that means my suggestion wouldn't work.

Issue #568, opened 2021-04-21
Can't get OAUTH via browser popup 


JorgeMiguelGomes:
### Problem

After revoking access to the rtweet app on one account, rtweet doesn't allow me to start the process again 

### Expected behavior

It would be expected that if no authentication is present then the popup would appear to allow access to rtweet via a twitter account 

### Reproduce the problem

``` r
## insert code here
library(rtweet)
algo <- lookup_users("jmgvostpt")

### rtweet version

<!-- run the code below and copy/paste the output -->

``` r
## copy/paste output
packageVersion("rtweet")
```
‘0.7.0’

### Session info

``` r
## copy/paste output
sessionInfo()
```
R version 4.0.5 (2021-03-31)
Platform: x86_64-apple-darwin17.0 (64-bit)
Running under: macOS Big Sur 10.16

Matrix products: default
LAPACK: /Library/Frameworks/R.framework/Versions/4.0/Resources/lib/libRlapack.dylib

locale:
[1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

loaded via a namespace (and not attached):
[1] compiler_4.0.5 tools_4.0.5    yaml_2.2.1     tinytex_0.31   xfun_0.22

llrs:
What do you mean by "doesn't allow me to start the process again"? Which error message do you get?

I think the problem is that the token is still saved on your computer and it is not automatically removed for you. You'll need to manually remove it and then rtweet will ask you to authenticate again. 
Let me know how it goes.

JorgeMiguelGomes:
> What do you mean by "doesn't allow me to start the process again"? Which error message do you get?
I don't  get any error message at all. 

> I think the problem is that the token is still saved on your computer and it is not automatically removed for you. You'll need to manually remove it and then rtweet will ask you to authenticate again.
> Let me know how it goes.

Sorry for the inconvenience but how do I do that?

JorgeMiguelGomes:
> Sorry for the inconvenience but how do I do that?

I can't find the .httr-oauth file on my system after running ```sudo find . -type f -name ".httr-ouath"``` @llrs

llrs:
Are you saying that if you do:

```r 
library(rtweet)
algo <- lookup_users("jmgvostpt")
```
You don't get any error? But you get something on algo?  
[...]

PR #591, opened 2021-06-28
Use tools::user_dir to store the configuration

llrs:
This is to comply with CRAN policies


hadley:
rappdirs already conforms to CRAN policies, so this shouldn't be necessary.

llrs:
Given that the next release will break where tokens are stored I prefer to use the new standard way of R to do it. One less direct dependency and probably more safe on the CRAN side.

hadley:
Your proposal means that only people with very recent versions of R will be able to use rtweet. It's your choice, but generally I think it's better to ensure that packages are usable by greatest number of people. The risk of using rappdirs is very low — it's had a total of 3 releases in its ~8 year life time, and since we use it in a number of tidyverse packages we will always adapt it to comply with changes in CRAN policies.

llrs:
I don't think a problem with the package is a lack of users. While I don't know the R version they use, maintaining the backward compatibility comes at the cost of my time maintaining the package, which is not part of my job. 

It is not a problem of risk on the rappdirs side but to unify the place where the packages will store user data from now on. However, I will make tools::user_dir conditional on the version used and provide some helpers to move tokens between previously used sites to store tokens: from TWITTERPATH, HOME and rappdirs path to the tools::user_dir path.

hadley:
I don't think it would be a good idea to make it more complicated. I'd suggest either leaving it as is (which works for all R versions and IMO is very low risk), or implementing this PR and deliberately breaking compatibility with older versions of R.

llrs:
There are [some reports](https://github.com/ropensci/rtweet/issues/568#issuecomment-824089360) of problems having multiple tokens stored on their computer. 
The CRAN policy says "the contents are actively managed (including removing outdated material)". Using rappdirs or tools::user_dir does not actively manage contents. 

My own experience managing/adding tokens to rtweet makes me think it is worth it to provide some helpers. Why do you consider this a bad idea?

hadley:
Are there other reports of problems? I think it would be a good idea to start by making precise exactly what the problems are before starting on a new approach. I'm quite aware of the CRAN policies around these directories and have used rappdirs without problems in a number of other packages.

llrs:
Yes, people [suggest to create a token on each session/code](https://github.com/ropensci/rtweet/issues/418#issuecomment-653481737) to prevent issues, and doing so [solved the issue](https://github.com/ropensci/rtweet/issues/418#issuecomment-653528339) for many people, as they upvoted [a comment](https://github.com/ropensci/rtweet/issues/418#issuecomment-694578510) indicating the same. 

The problems that need to be addressed under the actively managing material policy are:
 - Preventing to generate a token with the same keys for the same application. At one point I had 13 rtweet tokens stored on my computer, many of them were duplicates. 
 - Having tokens for rtweet at multiple locations.
 - Having to regenerate the token for an application upon next update of rtweet, instead of it moving the saved tokens. (This one is less of a concern for the default token but more for other tokens users might have saved). 

I'm glad you hadn't had any problem with CRAN. I'm not so confident I won't have any problem on CRAN. I think their policy to manage data stored is still useful. How do you remove outdated material with rappdirs on your other packages?

hadley:
Isn't that problem already fixed with the new approach in the development version? I thought we decided that there should be one default token that will be used by the majority of sessions.

llrs:
The current approach works well to set different tokens for different applications but doesn't solve any of the three problems I outlined above. How does the new development version manage multiple tokens on different paths of the users computer? How does rtweet currently remove duplicate tokens? 
With the new system to save tokens it is easier to avoid some of this problems and having a default token is useful for a single session. But managing and cleaning tokens for multiple sessions require more help from the package in my opinion.

hadley:
With the development version, if you don't want to use the standard rappdirs location, you have to mange the tokens yourself. I don't think there's anything else you can do here? And the chances of having a duplicate token should be much much lower, so I don't think it will be a scenario that needs so much consideration.

llrs:
Regardless of the location the three problems mentioned above are currently not managed by rtweet.
rtweet could have a function to check if an token is still active, remove duplicate tokens and move tokens of multiple paths. 

I understand that currently the scenario of having duplicate tokens is harder than on previous versions.  But CRAN doesn't use any conditional whether files should be actively managed.

llrs:
If no other comment I'll merge this soon (<1 week). There might be a hackathon and auth_sitrep might help anyone using the development version to manage their accounts (and it will be also a test to know if it is useful).

R/auth_sitrep.R
```diff
@@ -0,0 +1,187 @@
...
+#' @examples
+#' auth_sitrep()
+auth_sitrep <- function() {
+  all_tokens_files <- c(find_old_tokens(), 
+                        find_rappdirs_tokens(), 
+                        find_tools_tokens())
+  path <- tools::R_user_dir("rtweet", "config")
+  if (!dir.exists(path)) {
+    dir.create(path, showWarnings = FALSE, recursive = TRUE)
+  }
```
hadley, line 19: It's very hard to review this with all these codecov comments. Is there someway to suppress them?

hadley:
I'd really like to thoroughly review this PR before you merge it.

R/auth_sitrep.R
llrs: I wasn't even aware that this happened. I don't know when did they start appearing, but it is from the test-coverage workflow... I haven't found any way to change that (but haven't looked into the configuration of the workflow). Sorry!

llrs:
Ok, many thanks for providing feedback. I'll wait for your comments before doing something else on this branch.

R/auth_sitrep.R
llrs: I [found the way](https://docs.codecov.com/docs/common-recipe-list#set-non-blocking-status-checks,), I need to change codecov.yml to use informational: false. Will do that in this PR and on the other branches

hadley: Great, thanks!

R/auth.R
```diff
@@ -329,3 +323,13 @@ twitter_init_oauth1.0 <- function (endpoint, app, permission = NULL,
     private_key = private_key
   )
 }
+
+auth_path <- function(...) {
+  # Use private option to make testing easier
+  if (getRversion() >= "4.0.0") {
```
hadley, line 329: This conditioning means that different directories are used for different versions of R, e.g.:

``` r
tools::R_user_dir("rtweet", "config")
#> [1] "~/Library/Preferences/org.R-project.R/R/rtweet"
rappdirs::user_config_dir("rtweet")
#> [1] "~/Library/Application Support/rtweet"
```

This means that upgrading R might lose your token, and it'll be abandoned in a directory that is easy to forget about it. I think you need to pick one of the two options, rather than trying to support both.

R/auth_sitrep.R
```diff
@@ -0,0 +1,187 @@
+#' Twitter Tokens sitrep
+#' 
+#' Get a situation report of your current tokens; useful for upgrading from 
+#' rtweet 0.7.0 to 1.0.0 and diagnosing problems with tokens. Called by it side effects
```
hadley, line 4:
```suggestion
#' rtweet 0.7.0 to 1.0.0 and diagnosing problems with tokens. Called for its side effects
```

R/auth_sitrep.R
```diff
@@ -0,0 +1,187 @@
...
+#' 
+#' Searches for old tokens on the user folder, if duplicate tokens are found 
+#' they are deleted. Tokens are then moved to the new location. 
+#' @return Invisible the path of the tokens
```
hadley, line 8:
```suggestion
#' @return Invisibly, the path of the tokens
```

R/auth_sitrep.R
```diff
@@ -0,0 +1,187 @@
...
+#' @export
+#' @examples
+#' auth_sitrep()
+auth_sitrep <- function() {
+  all_tokens_files <- c(find_old_tokens(), 
+                        find_rappdirs_tokens(), 
+                        find_tools_tokens())
+  path <- tools::R_user_dir("rtweet", "config")
```
hadley, line 16: Shouldn't this use `auth_path()`?

R/auth_sitrep.R
```diff
@@ -0,0 +1,187 @@
...
+  if (!dir.exists(path)) {
+    dir.create(path, showWarnings = FALSE, recursive = TRUE)
+  }
+  
+  if (is.null(all_tokens_files)) {
+    inform("No tokens where found! See auth_as help file.")
```
hadley, line 22:
```suggestion
    inform("No tokens were found! See ?auth_as for more details.")
```

R/auth_sitrep.R
```diff
@@ -0,0 +1,187 @@
...
+#' rtweet 0.7.0 to 1.0.0 and diagnosing problems with tokens. Called by it side effects
+#' 
+#' Searches for old tokens on the user folder, if duplicate tokens are found 
+#' they are deleted. Tokens are then moved to the new location. 
+#' @return Invisible the path of the tokens
+#' @export
+#' @examples
+#' auth_sitrep()
+auth_sitrep <- function() {
+  all_tokens_files <- c(find_old_tokens(),
```
hadley, line 13: I think this line should move down below the directory creating code.

R/auth_sitrep.R
```diff
@@ -0,0 +1,187 @@
...
+  home_path <- normalizePath(file.path("~"), mustWork = TRUE)
+  
+  many_paths <- c(twitter_pat, home_path)
+  old_tokens <- lapply(many_paths, list.files, pattern = ".rtweet_token.*rds", 
+                       full.names = TRUE, all.files = TRUE)
+  unlist(old_tokens, TRUE, FALSE)
+}
+
+find_rappdirs_tokens <- function() {
+  if (requireNamespace("rappdirs", quietly = TRUE)) {
```
hadley, line 51: An early return would make this function a little easier to read:

```R
if (!is_installed("rappirs")) {
  return()
}

rappdirs_path <- ...
...
```

R/auth_sitrep.R
```diff
@@ -0,0 +1,187 @@
...
+  } else {
+    rappdirs_tokens <- NULL
+  }
+  rappdirs_tokens
+}
+
+find_tools_tokens <- function() {
+  path <- tools::R_user_dir("rtweet", "config")
+  final_path <- getOption("rtweet:::config_dir", path)
+  if (dir.exists(final_path)) {
```
hadley, line 65: You don't actually need this check here since list files will return an empty character vector if the path doesn't exist.

R/auth_sitrep.R
```diff
@@ -0,0 +1,187 @@
+#' Twitter Tokens sitrep
+#' 
+#' Get a situation report of your current tokens; useful for upgrading from
```
hadley, line 3: It appears to move the tokens around and create directories, which really doesn't fit into the framework of a "sitrep" function. I'd suggest a different name.

R/auth_sitrep.R
```diff
@@ -0,0 +1,187 @@
...
+
+#' @importFrom methods is
+type_auth <- function(old_tokens_files) {
+  old_tokens <- lapply(old_tokens_files, readRDS)
+  names(old_tokens) <- old_tokens_files
+  class_tokens <- sapply(old_tokens, is)
+  class_tokens <- ifelse(endsWith(class_tokens, "Token1.0"), "token", "bearer")
+}
+
+move_tokens <- function(tokens, folder) {
```
hadley, line 106: I worry that you have a lot of complicated code here and no tests. Tests are particularly important for base R file manipulating code, since most base R functions do not error if the operation fails (instead they warn and return a logical vector indicating whether or not each operation succeeded or failed).

llrs (commented):
Many thanks for all your comments. 
The supported path will become that provided by `tools:user_config_dir` and I will split the functionality between `auth_sitrep` to just report the status and auth_help(?) to fix some troubles users might have, as well as add some tests to check that this works correctly.

R/auth.R
hadley: Then you need to delete this code branch.

R/auth_sitrep.R
hadley: Yes, that would be a good idea.

llrs:
I'm not sure how to test moving files from the user .config folder without messing with CRAN checks or creating over-complicated test files, so I commented out this and left fixing the issues as an "exercise" for the user.

llrs:
If there is no new comment I'll merge this

PR #595, opened 2021-07-13
Allowing n=Inf in get_favorites (or elsewhere)

simonheb:
This is attempt to address https://github.com/ropensci/rtweet/issues/566.

The for-loop over pages is replaced by a while loop with a in-line counter. If `n` is finite the routine should be unchanged compared to the previous version. If `n` is finite, the result list is initiated to 1000 elements and doubled every time the boundary is hit (see: https://github.com/ropensci/rtweet/pull/567#issuecomment-821169712).

I followed @hadley's suggestion here, but I am not sure I understand why we cannot dynamically grow the return vector. The memory burden seems negligible to me, but I am clearly no expert here.

### DIFF
--- a/R/http.R
+++ b/R/http.R
@@ -121,10 +121,7 @@ TWIT_paginate_max_id <- function(token, api, params,
   params$since_id <- since_id
   params[[count_param]] <- page_size  
   pages <- ceiling(n / page_size)
-  if (is.finite(pages)) {
-    results <- vector("list", pages)
-  } else {
-    results <- vector("list", 1000) #see https://github.com/ropensci/rtweet/pull/567#issuecomment-821169712
+results <- vector("list", if (is.finite(pages)) pages else 1000)
   }
   
   if (verbose)  {
--- a/R/http.R
+++ b/R/http.R
@@ -121,12 +121,7 @@ TWIT_paginate_max_id <- function(token, api, params,
   params$since_id <- since_id
   params[[count_param]] <- page_size  
   pages <- ceiling(n / page_size)
-  
-  if (is.finite(n)) {
-    results <- vector("list", pages)
-  } else {
-    results <- vector("list", 2)
-  }
+  results <- vector("list", pages)
   
   if (verbose)  {
     pb <- progress::progress_bar$new(
@@ -135,10 +130,8 @@ TWIT_paginate_max_id <- function(token, api, params,
     ) 
     withr::defer(pb$terminate())
   }
-  
-  #because pages (because of n=Inf) might be ex ante infinite we can't use a for loop.
-  i <- 0 
-  while(i<-i+1 <=pages) {
+
+  for (i in seq_len(pages)) {
     params$max_id <- max_id
     if (i == pages) {
       params[[count_param]] <- n - (pages - 1) * page_size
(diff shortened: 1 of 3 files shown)


R/http.R
```diff
@@ -121,7 +121,11 @@ TWIT_paginate_max_id <- function(token, api, params,
   params$since_id <- since_id
   params[[count_param]] <- page_size  
   pages <- ceiling(n / page_size)
-  results <- vector("list", pages)
+  if (is.finite(pages)) {
+    results <- vector("list", pages)
+  } else {
+    results <- vector("list", 1000) #see https://github.com/ropensci/rtweet/pull/567#issuecomment-821169712
```
hadley, line 127:
```suggestion
results <- vector("list", if (is.finite(pages)) pages else 1000)
```

R/http.R
```diff
@@ -130,8 +134,9 @@ TWIT_paginate_max_id <- function(token, api, params,
     ) 
     withr::defer(pb$terminate())
   }
-
-  for (i in seq_len(pages)) {
+  
+  i<-0
+  while ((i<-i+1) <= pages) {
```
hadley, line 139: I think it's easier to understand if you do the check and the increment separately.

R/http.R
```diff
@@ -130,8 +134,9 @@ TWIT_paginate_max_id <- function(token, api, params,
     ) 
     withr::defer(pb$terminate())
   }
-
-  for (i in seq_len(pages)) {
+  
+  i<-0
```
hadley, line 138: Missing spaces, here and elsewhere

R/http.R
```diff
@@ -165,8 +173,7 @@ TWIT_paginate_max_id <- function(token, api, params,
       pb$tick()
     }
   }
-
-  results
+  results[1:i]
```
hadley, line 176: You should check that this is needed to avoid a potentially expensive copying of the list.

R/http.R
```diff
@@ -157,6 +162,9 @@ TWIT_paginate_max_id <- function(token, api, params,
     if (length(id) == 0) {
       break
     }
+    if(i>length(results)) { #doubling size, based on https://github.com/ropensci/rtweet/pull/567#issuecomment-821169712
```
hadley, line 165: I'd recommend commenting with a link to https://en.wikipedia.org/wiki/Dynamic_array#Geometric_expansion_and_amortized_cost or similar, which explains why this approach is needed.

hadley (commented):
Thanks for working on this!

R/http.R
simonheb: thanks for pointing this out. fixed.

R/http.R
simonheb: Yes. Turns out this change has had no effect anyways and this works also without it.

R/http.R
simonheb: yes, probably easier. changed this.

R/http.R
simonheb: thanks for the suggestion. did this.

R/http.R
```diff
@@ -157,6 +159,9 @@ TWIT_paginate_max_id <- function(token, api, params,
     if (length(id) == 0) {
       break
     }
+    if(i>length(results)) { #doubling size, see https://en.wikipedia.org/wiki/Dynamic_array#Geometric_expansion_and_amortized_cost
+      length(results) <- 2 * length(results)
```
hadley, line 163:
```suggestion
    if(i > length(results)) { 
      # double length per https://en.wikipedia.org/wiki/Dynamic_array#Geometric_expansion_and_amortized_cost
      length(results) <- 2 * length(results)
```

llrs:
It is much simpler than what I tried. Thanks @simonheb if nothing else raises I will merge it shortly. Thanks!

PR #634, opened 2021-11-22
fix YAML

maelle:
@llrs is this the intended structure?

screenshot of the reference index produced by this PR

### DIFF
--- a/_pkgdown.yml
+++ b/_pkgdown.yml
@@ -24,16 +24,20 @@ reference:
   - trends_available
   - starts_with("tweet")
 - subtitle: Lists
+  desc: Manage lists.
   contents:
     - starts_with("lists")
 - subtitle: Search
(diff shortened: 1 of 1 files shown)


llrs:
Yes, many thanks! I thought there would be more difference between titles and subtitles but a description to each title and subtitle  might help on this.

maelle:
That might be a problem solved when rotemplate switches to BS5 :thinking:

llrs:
I'm guessing that BS5 is the theme used. It might be just my preference/opinion not sure if it needs a change or it is simply that I'm  more familiar with the spacing on [ggplot2](https://ggplot2.tidyverse.org/reference/index.html) which has images that space the functions.

maelle:
No, rotemplate doesn't use BS5 yet but it will soon.

Right, adding small icons could help too. :-)

Issue #308, opened 2019-01-30
get_friends / get_follows consistency


alexpghayes:
`get_friends()` returns a tibble with two columns, but `get_followers()` only returns a tibble with a single column. It would be nice if they both returned the same columns. Always returning the two columns `user` and `user_id` would make it easy to build graphs from edge lists.

alexpghayes:
Side note: if there's going to be a breaking change here, I would *love* if we could get explicit `from` and `to` column names rather than `user` and `user_id`.

Arf9999:
Not to be difficult, but to avoid some breaking changes, it would be better not to change any column names. If some reference is required, rather add columns such as "follower_of" or "friend_of" which provide reference to a particular user_id for network mapping if required.

alexpghayes:
On top of this it seems like `get_friends()` only returns a single column when the requested user doesn't follow anyone.

``` r
rtweet::get_friends("hadleywickham")
#> # A tibble: 274 x 2
#>    user          ids                
#>    <chr>         <chr>              
#>  1 hadleywickham 1344499345773752321
#>  2 hadleywickham 3959153969         
#>  3 hadleywickham 43875304           
#>  4 hadleywickham 2423861950         
#>  5 hadleywickham 793171723772395521 
#>  6 hadleywickham 1218534623959044096
#>  7 hadleywickham 2798914670         
#>  8 hadleywickham 319822761          
#>  9 hadleywickham 43470348           
#> 10 hadleywickham 326511843          
#> # … with 264 more rows

rtweet::get_friends("SmallBuStudio")
#> # A tibble: 1 x 1
#>   user         
#>   <chr>        
#> 1 SmallBuStudio
```

<sup>Created on 2021-07-15 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

llrs:
I like the idea of making them more consistent and it is an easy change, but I don't like to add new columns with the same information. 

It might be worth adding to get_followers a first column with the user as on `get_friends`, that way the second column will always be the `to` user and the first one the `from`. The column names can be confusing and people that might not pay attention might think that the user_id is the id of the user on the first column, so a name change might be useful. 

For the moment when a user has no friends the output will be a 0 x 2 tibble, for easy `rbind`.

For reference on 0.7.0 they returned this:

```r
> rtweet::get_friends("SmallBuStudio")
## A tibble: 0 × 0
> rtweet::get_friends("hadleywickham")
[...]

PR #636, opened 2021-11-25
Adjust naming convention for follower and friend columns 

Arf9999:
Adjust names of columns to ensure consistency and avoid confusion across rtweet tibbles. Key column returned by `get_friends` and `get_followers` is a list of user_ids, for context the linked account should be identified, plus it should be clear the direction of the relationship. 

In order to do this, I have renamed the "from_id" variable as "friend_of" and "follower_of" respectively. The key variable is named "user_id" in both cases.

### DIFF
--- a/R/followers.R
+++ b/R/followers.R
@@ -17,8 +17,8 @@
 #' more_users <- get_followers("KFC", cursor = users)
 #'
 #' }
-#' @return A tibble data frame with one column named "from_id" with the 
-#' followers and another one "to_id" with the user used as input.
+#' @return A tibble data frame with one column named "follower_of" with the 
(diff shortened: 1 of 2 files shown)


llrs:
This goes the opposite direction of the issue #308 and would prevent merging `get_friends` and `get_followers` results easily. 

On the development version lookup_tweet returns "id"  and "id_str", not "user_id", so this won't help to your main problem to merge data.frames. To adapt the code on  https://github.com/arf9999/rtweetXtras/blob/master/R/rtweetxtras_functions.R,  you can use `dplyr::join_*(..., by = c("id_str" = "to_id")`.

PR #731, opened 2022-08-25
fix message in rtweet_bot authentication

filipemsc:
Bot authentication is asking for access token twice instead of asking for access token and access secret


llrs:
Thanks for fixing this, I hope it didn't cause too much confusion.

sin:
I just came here to fix this, thanks!

PR #737, opened 2022-10-04
Update README.Rmd

rmvegasm:
slight changes to the example on `get_timeline()` so that the heading, comment and code say the same thing


rmvegasm:
Thanks!
amazing package ;)

Issue #763, opened 2023-02-15
Call for co-maintainer


llrs:
I am looking for a co-maintainer to help me:
 - Supporting new endpoints, using httr2, testing an API in CI via httptest2, ...
 - Review changes to avoid new bugs, such as #756.
 - Help with issues and questions that the transition to API v2 and the current uncertainty around Twitter API.

I can help you with all the testing infrastructure (vcr, API credentials), github actions and the process of supporting and developing new functions. 
Of course, I will continue to answer issues and questions too, and you can ask me when in doubt.
For further context I [wrote a post](https://llrs.dev/post/2023/02/16/rtweet-future/).

I will take care of the CRAN submission process :smile:

PythonCoderUnicorn:
as someone who really enjoyed using this package before Musk,  I am interested, but have no experience on package maintenance and/or the specific help wanted points. 

Thanks for working on this package.

llrs:
Thanks! That's Ok, there is no need to put in X hours or do Z tasks. 
There is a lot that can be done without technical experience, for example just testing and reporting if the way the functions work make analysis simple would be helpful.

You can start looking at the [devel branch](https://github.com/ropensci/rtweet/tree/devel) to familiarize yourself with the code. I'll open pull requests for you to have a look (but I'm in the middle of big changes for the API v2 and I don't want to throw you under the bus). If you want, we can talk privately (are you in the rOpenSci slack?) about further steps.

llrs:
@PythonCoderUnicorn Hi Zane. Is there a way to get in touch with you privately to set up a meeting? Could you send me an email (it is in the package description file)?

PythonCoderUnicorn:
I did have initial interest, but considering the chaos happening at Twitter, 

-  I morally can't support twitter nor Musk
- doubt that the platform will last much longer with less staff by the week

I do wish you luck on finding some help, sorry.

llrs:
Understood. Thanks for volunteering, though!

howardbaek:
Are you still looking for a co-maintainer? I'm interested.

llrs:
Hi Howard, the situation is a bit tricky. 

When I posted this I was about to release a new version to CRAN and anticipating more usage and questions. Since the Twitter policy changes I have observed the opposite: less people using rtweet (but this might be a sampling error), and more questions about Twitter policies instead of rtweet. 

The API changes also made harder to test and check different endpoints and extend rtweet. 
[...]

PR #765, opened 2023-03-16
Improve documentation of what is the role of scopes

llrs:
I think this PR is easy to digest. Could you give it a look @ccamara if you are still interested in being a co-maintainer of the package #763?

I would like to know if it is easy to understand how to use rtweet_oauth2 and the client.

You will need to install rtweet as `devtools::install_github("ropensci@oauth2_vignette", build_vignettes = TRUE)` and check the new section with `vignette("auth", "rtweet")`.


llrs:
Any update @ccamara? I want to release a version next week.

